PR #7, opened 2020-03-16
BDOG-688: Wire NPM commands into SBT

cjwebb:
Please see scaladoc on JavaScriptBuild.scala for more information.

A future piece of work for this is to automatically compile & update
the JavaScript when 'sbt run' is executed. At the moment, it
doesn't do this. There is an example of this in cato-frontend with gulp.


project/Npm.scala
```diff
@@ -0,0 +1,30 @@
...
+  def npmProcess(failureMessage: String)(base: File, args: String*): Int = {
+    val processBuilder = process(base, args: _*)
+    val exitValue = processBuilder.run().exitValue()
+    if (exitValue != 0) {
+      throw new Exception(failureMessage)
+    } else exitValue
+  }
+
+  private def process(base: File, args: String*): ProcessBuilder = {
+    // this hasn't actually been tested on windows, so it might not work :)
```
natcarey: If it hasn't been tested on Windows I'd suggest erroring on Windows environments - it makes it easier to chase down the problem.

natcarey:
NPM Tests exit with a success status when a test failure occurs.  To replicate this update the test at `app/uk/gov/hmrc/trackingconsentfrontend/assets/test/renderBanner.spec.ts` to `expect(queryByText(document.body, /Tell us whether you accept cookes/)).toBeTruthy()` then run `sbt "npm run test" && echo 'done'` and the error appears in the console followed by `done`.

This means that if it were used on Jenkins test failures won't be flagged up.

project/Npm.scala
cjwebb: I copied it from cato-frontend, so I'm assuming it does actually work - its just a caveat.

cjwebb:
@natcarey I have updated the Npm class. The build would actually have used the 'npmProcess' code path, so the exit code was checked. Have updated so both the build, and the arbitrary npm commands are checked though.

natcarey:
Awesome.  Thanks for the explanation.  My use-case works now.

PR #8, opened 2020-03-17
Store preference for users who Accept All

natcarey:
This stores the preference but doesn't read it back.

I've used a self-contained cookie library as it makes the unit tests manageable.  I tried to use JSDOM's `CookieJar` but the integration point is difficult to track down and reset.

A few decisions which need reviewing:

- I'm not using secure (https only) cookies, these should be used in production but wouldn't be supported in local development.
- I've set the expiry for 10 years, I'm keen on storing the datetime that the user set the preferences so we can use that to invalidate them when we want to.  The approach I've chosen means we can choose to ask again for the user's preference after a year if we want/need to but we're not limited to doing that for technical reasons.  Some of our users will use the site annually.
- I put the event handler into the render function, I feel like that's not the best place for it.

A few changes I made alongside this:

- I added a reset to the render test, the previous version wasn't suitable for more than one test case.  This version is not future-proof, the reset only deals with elements added and not (things like) events on the body element which might cause problems later.
- I changed the lint task to *just* lint and not fix, to fix you'd need to run `npm run fixlint`.  I don't think we should be mutating code on a commit command.

The next step to hook this in is:

- To fire an event on the `datalayer` when we pick up the cookie which allows us to initialise tracking.  This should happen on each page load.

### DIFF
--- a/app/uk/gov/hmrc/trackingconsentfrontend/assets/package.json
+++ b/app/uk/gov/hmrc/trackingconsentfrontend/assets/package.json
@@ -8,7 +8,8 @@
   "scripts": {
     "start": "NODE_ENV=develop webpack-dev-server --config ./webpack.config.js --mode development",
     "build": "webpack --mode production",
-    "lint": "stylelint --fix \"**/*.scss\" && standardx --fix **/*.ts",
+    "lint": "stylelint \"**/*.scss\" && standardx **/*.ts",
+    "fixlint": "stylelint --fix \"**/*.scss\" && standardx --fix **/*.ts",
     "test": "jest --collectCoverage",
     "analyze": "webpack --mode production --profile --json > stats.json && webpack-bundle-analyzer stats.json"
   },
--- a/app/uk/gov/hmrc/trackingconsentfrontend/assets/package.json
+++ b/app/uk/gov/hmrc/trackingconsentfrontend/assets/package.json
@@ -54,6 +54,7 @@
     "identity-obj-proxy": "^3.0.0",
     "jest": "^25.1.0",
     "jest-environment-jsdom-fourteen": "^1.0.1",
+    "js-cookie": "^2.2.1",
     "node-sass": "^4.13.1",
     "sass-loader": "^8.0.2",
     "standardx": "^5.0.0",
@@ -64,9 +65,9 @@
     "ts-jest": "^25.2.1",
     "ts-loader": "^6.2.1",
     "typescript": "^3.8.3",
+    "uglifyjs-webpack-plugin": "^2.2.0",
     "webpack": "^4.42.0",
     "webpack-bundle-analyzer": "^3.6.0",
-    "uglifyjs-webpack-plugin": "^2.2.0",
     "webpack-cli": "^3.3.11",
     "webpack-dev-server": "^3.10.3"
   }
--- a/app/uk/gov/hmrc/trackingconsentfrontend/assets/src/ui/renderBanner.ts
+++ b/app/uk/gov/hmrc/trackingconsentfrontend/assets/src/ui/renderBanner.ts
@@ -1,10 +1,16 @@
 // @ts-ignore
 import bannerHtml from './banner.html';
 
-const renderBanner = () => {
+const renderBanner = (userPreference) => {
     const banner = document.createElement('div')
     banner.className = banner.className + ' cookie-banner';
     banner.innerHTML = bannerHtml
+    banner.querySelectorAll('.acceptAll').forEach(acceptAllButton => {
+        (acceptAllButton).addEventListener('click', e => {
+            e.preventDefault()
+            userPreference.userAcceptsAll()
+        })
+    })
     const parentNode = document.body
     parentNode.insertBefore(banner, parentNode.firstChild)
 }
--- a/app/uk/gov/hmrc/trackingconsentfrontend/assets/src/userPreferenceFactory.ts
+++ b/app/uk/gov/hmrc/trackingconsentfrontend/assets/src/userPreferenceFactory.ts
@@ -1,0 +1,16 @@
+// @ts-ignore
+import Cookies from 'js-cookie'
+
+const userPreferenceFactory = () => ({
+  userAcceptsAll: () => {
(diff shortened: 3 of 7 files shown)


app/uk/gov/hmrc/trackingconsentfrontend/assets/test/renderBanner.spec.ts
```diff
@@ -1,13 +1,46 @@
...
+    elem.dispatchEvent(event)
+  }
+  const assume = expect
+
+  beforeAll(reset)
+  afterEach(reset)
+
   it('should render a banner', () => {
-    renderBanner()
+    renderBanner(userPreferenceFactory())
```
matthewmascord: Given we're using a factory pattern, would it make sense to use it and pass in a stub rather than having to use the real implementation and spy on it? Otherwise what value is this indirection giving us?

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/ui/renderBanner.ts
```diff
@@ -1,10 +1,16 @@
 // @ts-ignore
 import bannerHtml from './banner.html';
 
-const renderBanner = () => {
+const renderBanner = (userPreference) => {
     const banner = document.createElement('div')
     banner.className = banner.className + ' cookie-banner';
     banner.innerHTML = bannerHtml
+    banner.querySelectorAll('.acceptAll').forEach(acceptAllButton => {
+        (acceptAllButton).addEventListener('click', e => {
```
matthewmascord: This won't work on IE8 :-) We should create a utility that determines whether to use attachEvent or addEventListener. Same applies to the DOM ready stuff in index.ts.

app/uk/gov/hmrc/trackingconsentfrontend/assets/package.json
```diff
@@ -8,7 +8,8 @@
   "scripts": {
     "start": "NODE_ENV=develop webpack-dev-server --config ./webpack.config.js --mode development",
     "build": "webpack --mode production",
-    "lint": "stylelint --fix \"**/*.scss\" && standardx --fix **/*.ts",
+    "lint": "stylelint \"**/*.scss\" && standardx **/*.ts",
+    "fixlint": "stylelint --fix \"**/*.scss\" && standardx --fix **/*.ts",
```
matthewmascord: Lint fix shouldn't change the semantics of the code, just style issues and formatting. I personally think it's fine to run as a pre-commit hook.

app/uk/gov/hmrc/trackingconsentfrontend/assets/test/renderBanner.spec.ts
```diff
@@ -1,13 +1,46 @@
...
   })
+
+  it('should call preference manager when user accepts', () => {
+    const userPreference = userPreferenceFactory()
+    spyOn(userPreference, 'userAcceptsAll')
+    renderBanner(userPreference)
+    const acceptAllButton = document.querySelector('.cookie-banner .acceptAll')
+    assume(userPreference.userAcceptsAll).not.toHaveBeenCalled()
+
+    click(acceptAllButton)
```
matthewmascord: DOM testing library has utility methods for this https://testing-library.com/docs/dom-testing-library/api-events

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/userPreferenceFactory.ts
```diff
@@ -0,0 +1,16 @@
+// @ts-ignore
+import Cookies from 'js-cookie'
+
+const userPreferenceFactory = () => ({
+  userAcceptsAll: () => {
+    Cookies.set('userConsent', JSON.stringify({
+      version: '2020-01-01',
```
cjwebb: Is 2020-01-01 the right value? It is March at the moment :)

app/uk/gov/hmrc/trackingconsentfrontend/assets/package.json
natcarey: So I'd argue strongly for not mutating code in any way during a commit.

The actual behaviour previously was that it mutated the code, the lint passed as a result but the changes made weren't included on the commit - therefore the commit itself had lint errors on it and I was left changes "not staged for commit".  If the goal is to keep the commit clean then running fix at the same time doesn't achieve that.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/ui/renderBanner.ts
natcarey: Fair, this morning's meeting sounds like we don't have to worry about that anymore.  I'd say we come back and fix these things if we start supporting IE8 again.

(for context - at the time you wrote that you're absolutely right and I shouldn't have overlooked it)

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/userPreferenceFactory.ts
natcarey: I'm happy either way.  I did change the interface slightly from the original `2020-01-01` so I'll update to `2020-03-01`.

app/uk/gov/hmrc/trackingconsentfrontend/assets/test/renderBanner.spec.ts
natcarey: That's fair.  I'll update.

app/uk/gov/hmrc/trackingconsentfrontend/assets/test/renderBanner.spec.ts
natcarey: Nice.

app/uk/gov/hmrc/trackingconsentfrontend/assets/package.json
matthewmascord: We could use lint-staged https://medium.com/@okonetchnikov/make-linting-great-again-f3890e1ad6b8#.8qepn2b5l - this would also make the linting process faster. But I understand if you feel strongly about not mutating before committing.

PR #11, opened 2020-03-18
Add cookie settings page

matthewmascord:
Stripped out hello world
Stripped out deprecated govuk-template and layout
Created new layout using play-frontend-govuk
Updated error template to use use play-frontend
Added content for cookie settings page
Add controller tests


app/uk/gov/hmrc/trackingconsentfrontend/controllers/CookieSettingsController.scala
```diff
@@ -17,22 +17,20 @@
...
 @Singleton
-class HelloWorldController @Inject()(appConfig: AppConfig, mcc: MessagesControllerComponents)
+class CookieSettingsController @Inject()(appConfig: AppConfig, mcc: MessagesControllerComponents, cookieSettings: CookieSettings)
     extends FrontendController(mcc) {
 
   implicit val config: AppConfig = appConfig
 
-  val helloWorld: Action[AnyContent] = Action.async { implicit request =>
-    Future.successful(Ok(hello_world()))
+  val cookie_settings: Action[AnyContent] = Action.async { implicit request =>
```
cjwebb: camelCase is the default for Scala. Please don't use underscores, as it makes me think there must be something special going on! :)
If you're concerned about name clash, I'd recommend changing the CookieSettings class to be called CookieSettingsPage - that's the pattern we've used in the Catalogue too.

conf/messages
```diff
@@ -0,0 +1,2 @@
+cookieSettings.title = Cookie settings on HMRC services
+service.name =
```
cjwebb: Is something missing here? service.name = what?

conf/app.routes
```diff
@@ -1,5 +1,6 @@
 # microservice specific routes
 
-GET        /hello-world             uk.gov.hmrc.trackingconsentfrontend.controllers.HelloWorldController.helloWorld
+->         /govuk-frontend          govuk.Routes
```
cjwebb: I don't really understand what this does? Have you got some docs somewhere about govuk frontendy stuff?

app/uk/gov/hmrc/trackingconsentfrontend/views/ErrorTemplate.scala.html
```diff
@@ -15,14 +15,14 @@
...
+@this(
+ layout: Layout,
+ paragraph: Paragraph,
+ h1: H1
+)
+@(pageTitle: String, heading: String, message: String)(implicit request: Request[_], messages: Messages, appConfig: AppConfig)
+@layout(pageTitle = Some(pageTitle)) {
+ @h1(content = heading)
+ @paragraph(content = Html(message))
+}
```
cjwebb: Are you anticipating re-using these views? Having separate templates for h1 and paragraph seems a little odd as they're only used once so far.

You can also set reusable snippets inside the same template like this: https://www.playframework.com/documentation/2.8.x/ScalaTemplates#Declaring-reusable-blocks

conf/app.routes
matthewmascord: This defines some routes for the govuk frontend assets: https://github.com/hmrc/play-frontend-govuk

app/uk/gov/hmrc/trackingconsentfrontend/views/ErrorTemplate.scala.html
matthewmascord: Yes, I have a follow-on PR with the static content that uses them.

conf/messages
matthewmascord: This suppresses the inclusion of a service name in the GovUK header given it's not service specific - @natcarey is there a better way to do this?

app/uk/gov/hmrc/trackingconsentfrontend/controllers/CookieSettingsController.scala
matthewmascord: Excellent, I much prefer camel case :-)

app/uk/gov/hmrc/trackingconsentfrontend/views/ErrorTemplate.scala.html
matthewmascord: They are also used by the error template.

PR #14, opened 2020-03-19
BDOG-705 user revoke consent (part 2)

natcarey:
In this PR we have:

 - Moved all copy into the message file
 - Established the interface between the page and the Javascript
 - Bundled a JS file specifically for this page (which can rely on modules like `userPreferenceFactory`

We still need to:

 - Hook up the radio buttons as input and output


cjwebb:
Looks good

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/cookieSettingsPage.ts
```diff
@@ -0,0 +1,9 @@
+const castToArray = nodeList => Array.prototype.slice.apply(nodeList)
+
+document.addEventListener('DOMContentLoaded', () => {
+  castToArray(document.querySelectorAll('[data-module="cookie-settings"]')).forEach(cookieSettingsForm => {
```
matthewmascord: You could use the array spread syntax here - it should be transpiled by Babel.

matthewmascord: i.e. [...document.querySelectorAll('[data-module="cookie-settings"]')] - I could be wrong!

natcarey: I just tried replacing `const castToArray = nodeList => Array.prototype.slice.apply(nodeList)` with `const castToArray = nodeList => [...nodeList]` and it's being simplified down to `[].concat(nodeList)` which just wraps my nodeList in an array.

matthewmascord: Weird, it should work.

PR #18, opened 2020-03-20
BDOG-708 Add test endpoint

matthewmascord:
Add a test endpoint to demonstrate use of the cookie banner partial injection on Play 2.6.

CookieModule (from @cjwebb)

This wires up the CookieBanner for Guice injection. It is intended that
this code will be put into bootstrap-play-26 at some point soon, which
means that service teams will have less to do.

If service teams cannot use the latest bootstrap, they will need to
duplicate this bit of code. It will be included in the documentation.


cjwebb:
Is it worth having a test for the test controller?

PR #23, opened 2020-03-24
BDOG-726: Update port, and move test endpoint route

cjwebb:


conf/testOnlyDoNotUseInAppConf.routes
```diff
@@ -11,3 +11,4 @@
 
 # Add all the application routes to the prod.routes file
 ->         /                          prod.Routes
+GET        /test-only/                uk.gov.hmrc.trackingconsentfrontend.controllers.TestController.test
```
matthewmascord: Will this still work in QA?

conf/application.conf
```diff
@@ -109,5 +109,5 @@ assets {
 }
 
 cookie-banner {
-    port = 9000
+    port = 12345
```
matthewmascord: This has actually just gone in already.

conf/testOnlyDoNotUseInAppConf.routes
cjwebb: Its configurable with the `application.router` config property.
Do we need it to work in QA?

matthewmascord: Well yes ideally, for demonstration purposes?

PR #24, opened 2020-03-24
Add additional test cases and handle malformed data

matthewmascord:
Ignore data the wrong shape
Ignore data with the wrong types
Increase coverage threshold to 100%

### DIFF
--- a/app/uk/gov/hmrc/trackingconsentfrontend/assets/src/domain/userPreferenceFactory.ts
+++ b/app/uk/gov/hmrc/trackingconsentfrontend/assets/src/domain/userPreferenceFactory.ts
@@ -2,7 +2,7 @@
 import Cookies from 'js-cookie'
 import cookieTypes from '../constants/cookieTypes'
 import fromEntries from '../common/fromEntries'
-import { COOKIE_CONSENT } from '../constants/cookies'
+import {COOKIE_CONSENT, COOKIE_VERSION} from '../constants/cookies'
 
 function storePreferences (preferences) {
   Cookies.set(COOKIE_CONSENT, {
@@ -26,16 +26,16 @@ const setPreferences = preferencesIn => {
   })
 }
 
-const sanitizePreferences = preferences => {
-  const entries: [string, boolean][] = cookieTypes.reduce((accumulator, cookieType) => {
+const isPreferenceStrictlyBoolean = ([, value]) => typeof value === 'boolean'
+
+const getSanitisedPreferences = preferences => {
+  const entries: [string, boolean][] = cookieTypes.map(cookieType => {
     const { [cookieType]: value } = preferences
 
-    // If value is strictly not boolean, then do not assume either way
-    // Otherwise, return boolean as-is
-    return typeof value !== 'boolean' ? accumulator : [...accumulator, [cookieType, value]]
+    return [cookieType, value]
   }, [])
 
-  return fromEntries(entries)
+  return entries.filter(isPreferenceStrictlyBoolean)
 }
 
 const allThePreferences = () => fromEntries(cookieTypes.map(cookieType => [cookieType, true]))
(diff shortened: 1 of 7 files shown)


app/uk/gov/hmrc/trackingconsentfrontend/assets/src/domain/userPreferenceFactory.ts
```diff
@@ -23,26 +26,36 @@ const setPreferences = preferencesIn => {
   })
 }
 
+const sanitizePreferences = preferences => {
```
natcarey: We tend to avoid Americanisms, `sanitisePreferences` would be more in keeping with other projects I've worked on.  (for example GDS use `colour` constants rather than `color`.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/domain/userPreferenceFactory.ts
```diff
@@ -23,26 +26,36 @@ const setPreferences = preferencesIn => {
...
+  const entries: [string, boolean][] = cookieTypes.reduce((accumulator, cookieType) => {
+    const { [cookieType]: value } = preferences
+
+    // If value is strictly not boolean, then do not assume either way
```
natcarey: Is there a way of making the code read clearly enough without a comment?  Would wrapping this in a well named function work?

app/uk/gov/hmrc/trackingconsentfrontend/assets/test/domain/userPreference.spec.ts
```diff
@@ -74,11 +73,53 @@ describe('User Preference Factory', () => {
...
+
+      expect(userPreferenceFactory().getPreferences()).toEqual({ usage: true })
+    })
+
+    it('should ignore any undefined values', () => {
+      setConsentCookie({ preferences: { usage: true, settings: undefined } })
+
+      expect(userPreferenceFactory().getPreferences()).toEqual({ usage: true })
+    })
+
```
natcarey: Can we add a case for an unexpected version please?  `version: 2020-03-01` should be the only value we accept.

app/uk/gov/hmrc/trackingconsentfrontend/assets/test/domain/userPreference.spec.ts
```diff
@@ -88,6 +129,17 @@ describe('User Preference Factory', () => {
...
     })
+
+    it('should not return each category when AcceptAll is non-boolean but truthy', () => {
+      setConsentCookie({
+        preferences: {
+          acceptAll: 'abc'
+        }
+      })
+
+      expect(userPreferenceFactory().getPreferences()).toEqual({})
```
natcarey: Should this be returning `undefined`?  Here's what I'd expect:

`{ usage: true }` -> `{ usage: true }`
`{ usage: true, settings: 'abc' }` -> `{ usage: true }` (unacceptable value ignored)
`{ usage: true, someNewThing: true }` -> `{ usage: true }` (unknown key ignored)
`{ usage: 'abc', someNewThing: true }` -> `undefined` (unacceptable value ignored, unknown key ignored, no preferences to return)

Essentially the approach I'm suggesting is to treat unreadable answers as though they didn't exist, if after that nothing exists then `undefined` is a good thing to return.

app/uk/gov/hmrc/trackingconsentfrontend/assets/test/domain/userPreference.spec.ts
```diff
@@ -195,9 +235,9 @@ describe('User Preference Factory', () => {
       expect(new Date().getTime()).toEqual(123)
     })
     it('cookies should be settable and gettable', () => {
-      expect(Cookies.get('abcdef')).toBeUndefined()
+      expect(Cookies.getJSON('abcdef')).toBeUndefined()
       Cookies.set('abcdef', 'this is my cookie value')
-      expect(Cookies.get('abcdef')).toBe('this is my cookie value')
+      expect(Cookies.getJSON('abcdef')).toBe('this is my cookie value')
```
natcarey: This is to prove that the mock behaves correctly, it would be good to show it working with an object.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/domain/userPreferenceFactory.ts
natcarey: I wonder whether this would read well as a filter.  `.filter(booleanValues)`, `.filter(strictBooleanValues)` or `.filter(keepBooleanValues)`.

PR #26, opened 2020-03-25
BDOG 717: Add cookie settings save confirmation

matthewmascord:
Add save confirmation
Renamed bundle
Add link to referring page as per gov.uk implementation
Add scroll to top of page
Fix typo on settings page radio


app/uk/gov/hmrc/trackingconsentfrontend/assets/test/common/getReferrer.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+
+        expect(referrer).toEqual('/another-service')
+    })
+
+    it('should return an empty string if no referrer has been set', () => {
+        documentReferrer = ''
+        const referrer = getReferrer()
+
+        expect(referrer).toEqual('')
+    })
```
cjwebb: More of a general query about conventions, but why is empty string preferred to null here?
I see there's an `if(referrer)` check elsewhere, so the behavior would be the same... I was just wondering, as I'm not too familiar with conventions that people use for JS/TS.

matthewmascord: Good question. I guess I was going for consistency with https://developer.mozilla.org/en-US/docs/Web/API/Document/referrer

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/ui/renderSettingsSaveConfirmationMessage.ts
```diff
@@ -0,0 +1,45 @@
...
+import scrollToTop from "../common/scrollToTop";
+import getReferrer from "../common/getReferrer";
+import {GO_BACK_TO_PREVIOUS_PAGE} from "../constants/messages";
+
+const insertReferrerLink = (notice, referrer) => {
+    const link = document.createElement('a')
+    link.className = GOV_UK_LINK
+    link.innerHTML = GO_BACK_TO_PREVIOUS_PAGE
+    link.href = referrer
+    notice.appendChild(link)
```
natcarey: Links should be wrapped in paragraph tags with `govuk-body` class.

This is one of the times where a templating language would make this easier/better.  I don't think it's a strong enough case at the moment but it's something to watch for.  It would be nice to be able to render a success banner and put an optional back link into the model.

Another approach for this could be to put the link into `cookieSettingsConfirmation.html`, then find it and either set the `href` or delete it from the DOM depending on whether we have the link.

This seems to be the first time that JS/TS is in control of the copy, I'd rather avoid that if it's not disproportionately difficult.

app/uk/gov/hmrc/trackingconsentfrontend/assets/test/ui/renderSettingsSaveConfirmationMessage.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+        // @ts-ignore
+        expect(container.classList).toContain('cookie-settings__notice')
+    })
+
+    it('should not render the referrer link if not set (to be consistent with Gov.UK implementation)', () => {
+        referrer = ''
+        renderSettingsSaveConfirmationMessage()
+
+        expect(queryAllByText(document.body, /Go back to the page you were looking at/).length).toEqual(0)
+    })
```
natcarey: It would be good to add a test to show that the page scrolls up every time even though the message is only added once.

app/uk/gov/hmrc/trackingconsentfrontend/assets/test/ui/renderSettingsSaveConfirmationMessage.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+})
```
natcarey: Should the message be hidden if the user changes their response?  I feel like a page which is no longer in the state which was saved should not still have a save confirmation.

matthewmascord: It would be good to get UX input into that. The implementation I've gone with is consistent with the Gov.UK solution.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/ui/renderSettingsSaveConfirmationMessage.ts
matthewmascord: I was following the Gov.UK example that did not have the link wrapped in a <p>

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/ui/renderSettingsSaveConfirmationMessage.ts
```diff
@@ -0,0 +1,45 @@
...
+import confirmationHtml from './cookieSettingsConfirmation.html'
+import callIfNotNull from "../common/callIfNotNull";
+import scrollToTop from "../common/scrollToTop";
+import getReferrer from "../common/getReferrer";
+import {GO_BACK_TO_PREVIOUS_PAGE} from "../constants/messages";
+
+const insertReferrerLink = (notice, referrer) => {
+    const link = document.createElement('a')
+    link.className = GOV_UK_LINK
+    link.innerHTML = GO_BACK_TO_PREVIOUS_PAGE
```
natcarey: Should this be `innerText`?

matthewmascord: Would you prefer it to be?

natcarey: Yeah, it is text rather than HTML.

matthewmascord: Is innerText settable?

PR #28, opened 2020-03-27
BDOG-709 & BODG-722 triggering GTM events

natcarey:
This includes the interface for the communicator (currently communicating with GTM).

This PR also includes communicating with GTM when:

 - The user hits accept all
 - The preferences page is saved
 - Each page loads when the user has consented

### DIFF
--- a/app/uk/gov/hmrc/trackingconsentfrontend/assets/src/domain/userPreferenceFactory.ts
+++ b/app/uk/gov/hmrc/trackingconsentfrontend/assets/src/domain/userPreferenceFactory.ts
@@ -4,70 +4,80 @@ import cookieTypes from '../constants/cookieTypes'
 import fromEntries from '../common/fromEntries'
 import {COOKIE_CONSENT, COOKIE_VERSION} from '../constants/cookies'
 
-function storePreferences (preferences) {
-  Cookies.set(COOKIE_CONSENT, {
-    version: '2020-03-01',
-    dateSet: new Date().getTime(),
-    preferences: preferences
-  }, { sameSite: 'strict', expires: 3650 })
-}
 
-const userAcceptsAll = () => {
-  storePreferences({
-    acceptAll: true
-  })
-}
-
-const setPreferences = preferencesIn => {
-  storePreferences({
-    usage: preferencesIn.usage,
-    campaigns: preferencesIn.campaigns,
-    settings: preferencesIn.settings
-  })
-}
-
-const isPreferenceStrictlyBoolean = ([, value]) => typeof value === 'boolean'
-
-const getSanitisedPreferences = preferences => {
-  const entries: [string, boolean][] = cookieTypes.map(cookieType => {
-    const { [cookieType]: value } = preferences
-
-    return [cookieType, value]
-  }, [])
-
-  return entries.filter(isPreferenceStrictlyBoolean)
-}
-
-const allThePreferences = () => fromEntries(cookieTypes.map(cookieType => [cookieType, true]))
-
-const getPreferences = () => {
-  const cookie = Cookies.getJSON(COOKIE_CONSENT)
-  if (cookie === null || cookie === undefined || cookie.preferences === undefined) {
-    return undefined
+const userPreferenceFactory = (preferenceCommunicator) => {
+  function storePreferences (preferences) {
+    Cookies.set(COOKIE_CONSENT, {
+      version: '2020-03-01',
+      dateSet: new Date().getTime(),
+      preferences: preferences
+    }, { sameSite: 'strict', expires: 3650 })
   }
 
-  const { version } = cookie
-  if (version !== COOKIE_VERSION) {
-    return undefined
+  const userAcceptsAll = () => {
+    storePreferences({
+      acceptAll: true
+    })
+    preferenceCommunicator.sendPreferences(self)
   }
 
-  const { preferences } = cookie
-  if (preferences.acceptAll === true) {
-    return allThePreferences()
+  const setPreferences = preferencesIn => {
+    storePreferences({
+      usage: preferencesIn.usage,
+      campaigns: preferencesIn.campaigns,
+      settings: preferencesIn.settings
+    })
+    preferenceCommunicator.sendPreferences(self)
   }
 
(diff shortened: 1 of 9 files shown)


app/uk/gov/hmrc/trackingconsentfrontend/assets/src/interfaces/preferenceCommunicatorFactory.ts
```diff
@@ -0,0 +1,18 @@
+const preferenceCommunicatorFactory = (window) => ({
+  sendPreferences: (userPreferences) => {
+    window.dataLayer = window.dataLayer || []
+    const preferences = userPreferences.getPreferences() || {}
+    console.log(preferences)
```
matthewmascord: Console.log in code - surprised linter hasn't complained.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/ui/renderBanner.ts
```diff
@@ -18,20 +19,20 @@ const hideQuestion = () => {
     callIfNotNull(question, element => removeElement(question))
 }
 
-const insertBanner = (userPreference) => {
+const insertBanner = (userPreference, preferenceCommunicator) => {
```
matthewmascord: We should move these function declarations inside renderBanner so we don't have to pass through parameters like this. insertBanner shouldn't really care about preferenceCommunicator.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/ui/renderBanner.ts
```diff
@@ -5,10 +5,11 @@ import { COOKIE_BANNER_QUESTION_CLASS, COOKIE_BANNER_CLASS, ACCEPT_ALL_CLASS } f
 import removeElement from '../common/removeElement';
 import callIfNotNull from '../common/callIfNotNull';
 
-const handleAcceptAllClick = userPreference => event => {
+const handleAcceptAllClick = (userPreference, preferenceCommunicator) => event => {
     event.preventDefault()
 
     userPreference.userAcceptsAll()
+    preferenceCommunicator.sendPreferences(userPreference)
```
matthewmascord: It feels like this call should belong in the business logic layer - i.e. the userAcceptsAll method should be responsible for notifying GTM (possibly other systems in future - could adopt a pub/sub pattern) via the communicator. The UI layer should just have to call one thing.

app/uk/gov/hmrc/trackingconsentfrontend/assets/test/ui/renderBanner.spec.ts
```diff
@@ -42,22 +50,34 @@ describe('renderBanner', () => {
...
-    renderBanner(userPreference)
+    renderBanner(userPreference, preferenceCommunicator)
     assume(userAcceptsAll).not.toHaveBeenCalled()
 
     clickAcceptAll()
 
     expect(userAcceptsAll).toHaveBeenCalledWith()
   })
 
+  it('should call communicator after setting preferences', () => {
```
matthewmascord: Just re-iterating, might be worth considering moving this responsibility to the userPreference object. If we add more interfaces in future, the UI code is going to get polluted with concerns around managing the communication of preferences to other systems which I don't think is it's responsibility.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/interfaces/preferenceCommunicatorFactory.ts
natcarey: Doh, it would be good for the linter to complain over that.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/ui/renderBanner.ts
natcarey: I went through a few options on this one.  That was where I started but I moved away from it.  We need to send the preferences:

 - When saving settings
 - When accepting all
 - When page loads

The event option seems sensible but I'm not sure this is a big enough use-case to justify an event system.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/ui/renderBanner.ts
natcarey: Yeah, I think the event handler should be configured elsewhere.

app/uk/gov/hmrc/trackingconsentfrontend/assets/test/ui/renderBanner.spec.ts
natcarey: That's fair, I'll try moving the communicator function back into the `userPreferences` and see whether that fits better.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/domain/userPreferenceFactory.ts
```diff
@@ -4,70 +4,76 @@ import cookieTypes from '../constants/cookieTypes'
...
+
+    const sanitisedPreferences = getSanitisedPreferences(preferences)
+
+    return sanitisedPreferences.length > 0 ? fromEntries(sanitisedPreferences) : undefined
+  }
+
+  const getUserHasSavedCookiePreferences = () => getPreferences() !== undefined
 
-const getUserHasSavedCookiePreferences = () => getPreferences() !== undefined
+  const self = {
```
matthewmascord: What's the purpose of defining this?

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/interfaces/preferenceCommunicatorFactory.ts
```diff
@@ -0,0 +1,17 @@
+const preferenceCommunicatorFactory = (window) => ({
+  sendPreferences: (userPreferences) => {
```
matthewmascord: Could this not just take the actual preferences rather than the userPreferences object? Worth adding a type too.

PR #30, opened 2020-03-30
BDOG-715: Use govuk-frontend styles as much as possible

matthewmascord:
Removes bespoke styles and replaces with govuk-frontend styles as much as possible.
Imports govuk-frontend mixins, helpers and settings (pinned to version 3.6.0)
Clones and cuts down any needed global styles from govuk-frontend and prefixes with cbanner to avoid conflict with any style overrides set by services.
Updates markup to be consistent with govuk-frontend


app/uk/gov/hmrc/trackingconsentfrontend/assets/package.json
```diff
@@ -71,5 +72,8 @@
     "webpack-bundle-analyzer": "^3.6.0",
     "webpack-cli": "^3.3.11",
     "webpack-dev-server": "^3.10.3"
+  },
+  "dependencies": {
+    "govuk-frontend": "3.6.0"
```
natcarey: I'd say this is a dev dependency.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/constants/cssClasses.ts
```diff
@@ -1,11 +1,18 @@
...
+export const COOKIE_SETTINGS_NOTICE_CLASS = 'cookie-settings__notice'
 export const COOKIE_SETTINGS_CONFIRMATION_CLASS = 'cookie-settings__confirmation'
 export const COOKIE_SETTINGS_WRAPPER_CLASS = 'govuk-grid-column-two-thirds'
-export const ACCEPT_ALL_CLASS = 'acceptAll'
-export const HIDE_BUTTON_CLASS = 'cookie-banner__hide-button'
-export const GOV_UK_LINK = 'govuk-link'
-export const GOV_UK_BODY = 'govuk-body'
+
+// Gov UK frontend
+export const GOV_UK_BODY_CLASS = 'govuk-body'
```
natcarey: I'd assume these should be prefixed.

matthewmascord: These are the ones used on the cookie settings page, that are vanilla govuk-frontend.

natcarey:
I've explored alternatives.  I think PurgeCSS would be a reasonable way of using styles from GovUK if this approach starts to show limitations.  There's a proof of concept here https://gist.github.com/natcarey/83904cde27e64f27f6730996b9f1140d

PR #34, opened 2020-04-02
BDOG-769: Fix double GTM fire issue, refactor

matthewmascord:
Fixing issue discovered today relating to GTM firing twice. Also refactor code to remove duplicated logic and make render functions consistent.


app/uk/gov/hmrc/trackingconsentfrontend/assets/test/ui/renderBanner.spec.ts
```diff
@@ -59,13 +60,13 @@ describe('renderBanner', () => {
   }
 
   it('should render a banner', () => {
-    renderBanner(userPreference)
+    renderBanner(document, userPreference)
```
natcarey, line 63: It would be good to create a JSDOM for each test like https://github.com/hmrc/tracking-consent-frontend/blob/master/app/uk/gov/hmrc/trackingconsentfrontend/assets/test/ui/settingsFormHandler.spec.ts#L24 and remove the need for line 41 as a reset.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/ui/renderBanner.ts
```diff
@@ -50,10 +51,10 @@ const insertBanner = (userPreference: UserPreferences) => {
     insertAfterSkipLink(banner)
 }
 
-const renderBanner = (userPreference: UserPreferences) => {
+export const renderBanner = (_, userPreference: UserPreferences) => {
```
natcarey, line 54: Is underscore an unused `document`?  renderBanner uses `document` a few times, surely it should use the one that's provided to it?

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/common/withContentLoaded.ts
```diff
@@ -0,0 +1,7 @@
+const withContentLoaded = render => (document, ...args) => {
```
natcarey, line 1: I think this should have some tests around it.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/common/pageHandler.ts
```diff
@@ -0,0 +1,10 @@
+import enableGtm from "../interfaces/gtm";
+import {UserPreferences} from "../../types/UserPreferences";
+
+const pageHandler = (userPreference: UserPreferences, pageRenderer) => {
```
natcarey, line 4: I think this should have some tests around it.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/ui/renderBanner.ts
natcarey, line 54: Why is this being exported?  I'm assuming it's to test without the `withContentLoaded` condition needing to be met.  If we're taking that approach we should also be testing the default export as that's what other includes are using.

I would personally rather test it in the same way as the settings form is tested - firing the page load and seeing how the unit responds to it.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/common/pageHandler.ts
matthewmascord: There are tests for this in tests/common/pageHandler.spec.ts

app/uk/gov/hmrc/trackingconsentfrontend/assets/test/ui/renderBanner.spec.ts
matthewmascord: I think we should move away from creating a new JSDOM for each test - this is causing our tests to run very slowly.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/ui/renderBanner.ts
matthewmascord: I actually think it's unnecessary passing a document in. It is fundamentally a global. I added the param purely to be consistent with the formHandler.

matthewmascord: I would rather go the other way and just test the rendering part rather than with the DOM Content Loaded handling, as this functionality can be tested separately.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/common/withContentLoaded.ts
matthewmascord: I've added a test. We could consider removing the equivalent tests from the formHandler.

app/uk/gov/hmrc/trackingconsentfrontend/assets/test/ui/renderBanner.spec.ts
natcarey: In that case I think we should use a different mock/stub of document where we mock out the functions we need.  The shared scope between unit tests is really worth avoiding.

matthewmascord: OK, but this is not really relating to this PR as the tests haven't changed. Can we address this later on?

PR #41, opened 2020-05-21
PLATUI-465: Add content for unsupported browsers

matthewmascord:
This PR adds the appropriate content for users of unsupported browsers.

Also refactors tests to be consistent.

### DIFF
--- a/app/uk/gov/hmrc/trackingconsentfrontend/assets/src/ui/settingsFormHandler.ts
+++ b/app/uk/gov/hmrc/trackingconsentfrontend/assets/src/ui/settingsFormHandler.ts
@@ -61,7 +61,23 @@ const hydrateForm = (userPreferences: UserPreferences) => (form: HTMLFormElement
   form.addEventListener('submit', submitHandler)
 }
 
+const hideUnsupportedBrowserContent = () => {
+  const element = document.querySelector('.cookie-settings__unsupported-browser')
+  callIfNotNull(element, element => {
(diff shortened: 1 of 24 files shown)


natcarey:
This implementation covers:

 - Javascript-enabled browsers we know are unsupported 
 - content which is on the page when the detection runs

I'd suggest we take a different approach for the CSS rules:

- when we detect a supported browser we add a class to the body e.g. `tracking-consent-frontend-supported-browser`
 - We set up CSS rules to show unsupported content if that body class isn't present.

That also allows us to hide content which appears after the detection has run which should be more robust.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/ui/settingsFormHandler.ts
```diff
@@ -61,7 +61,12 @@ const hydrateForm = (userPreferences: UserPreferences) => (form: HTMLFormElement
   form.addEventListener('submit', submitHandler)
 }
 
+const hideUnsupportedBrowserContent = () => {
+  document.body.className = `${document.body.className} cookie-settings__body--browser-supported`
```
natcarey, line 65: Would it be better to `+=`?

matthewmascord: Ha - maybe. I've followed https://eslint.org/docs/rules/prefer-template in the past so would generally avoid concatenation and += generally. Like you say it's a matter of style.

PR #56, opened 2020-06-29
Platui 542 cookie banner feature toggle

JoPintoPaul:


app/uk/gov/hmrc/trackingconsentfrontend/assets/src/interfaces/isFeatureEnabled.ts
```diff
@@ -0,0 +1,22 @@
+import Cookies from 'js-cookie'
+
+const isFeatureEnabled = (featureName: String) => {
+    function getEnabledFlagFromCookie(featureName: String) {
```
matthewmascord, line 4: Could use arrow function const getEnabledFlatFromCookie = ... as above?

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/interfaces/isFeatureEnabled.ts
```diff
@@ -0,0 +1,22 @@
...
+        return (Cookies.get(featureName) == 'true')
+    }
+
+    function getEnabledFlagFromQuery(featureName: String) {
+        return window.location.search.includes(featureName.concat('=true'))
```
matthewmascord, line 9: Would an ES6 template literal be easier to read here vs .concat?

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/interfaces/isFeatureEnabled.ts
JoPintoPaul: Done.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/interfaces/isFeatureEnabled.ts
JoPintoPaul: Done.

PR #65, opened 2020-07-15
PLATUI-573: Add compression and reduce TTL of assets from 3600 seconds

matthewmascord:
Follow best practices for assets as per: https://www.playframework.com/documentation/2.7.x/AssetsOverview

Compress all assets using gzip
Reduce default TTL of assets from 3600 seconds to
- 60 seconds for tracking.js and tracking-transitional.js
- 0 seconds for the cookie settings page assets

In practice most requests will be using the Etag resulting in a 304 Not Modified response. Having a TTL > 0 for tracking.js will reduce the load on the server for this shared asset.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -65,6 +65,7 @@ lazy val microservice = Project(appName, file("."))
     ),
     PlayKeys.playRunHooks += Webpack(javaScriptDirectory.value),
     PlayKeys.devSettings ++= Seq("metrics.enabled"  -> "false", "auditing.enabled" -> "false"),
+    pipelineStages in Assets := Seq(gzip),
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -65,6 +65,7 @@ lazy val microservice = Project(appName, file("."))
     ),
     PlayKeys.playRunHooks += Webpack(javaScriptDirectory.value),
     PlayKeys.devSettings ++= Seq("metrics.enabled"  -> "false", "auditing.enabled" -> "false"),
+    pipelineStages in Assets := Seq(gzip),
```
natcarey, line 68: I'm wondering if this should be part of the bootstrap.

PR #77, opened 2020-08-25
PLATUI-611_resolve_open_redirect

matthewmascord:


app/uk/gov/hmrc/trackingconsentfrontend/config/AppConfig.scala
```diff
@@ -35,5 +35,12 @@ class AppConfig @Inject()(config: Configuration, servicesConfig: ServicesConfig)
...
       s"$baseUrl$projectId.js"
     }
-  val languages: Seq[String] = config.getOptional[Seq[String]]("play.i18n.langs").getOrElse(Seq("en"))
+
+  val en: String            = "en"
+  val cy: String            = "cy"
+
+  private val platformFrontendHost = config.getOptional[String]("platform.frontend.host")
+  val languageControllerHostUrl: String = platformFrontendHost.getOrElse(
+    config.get[String]("language-controller.host")
```
natcarey, line 44: What happens if this is not set?  Is there a danger that without this set we (still) allow redirects to any URL?

matthewmascord: config.get will throw an exception if not set, probably worth adding a test for this though.

natcarey (approved):

PR #78, opened 2020-08-25
PLATUI-549: Remove old banner

matthewmascord:


app/uk/gov/hmrc/trackingconsentfrontend/assets/src/ui/renderBanner.ts
```diff
@@ -52,7 +52,14 @@ const renderBanner = (userPreference: UserPreferences) => {
...
 
+    const removeLegacyCookieBanner = () => {
+        const legacyBanner = document.querySelector(`#${LEGACY_COOKIE_BANNER_ID}`)
+
+        callIfNotNull(legacyBanner, removeElement)
+    }
+
     if (!userPreference.getUserHasSavedCookiePreferences()) {
+        removeLegacyCookieBanner()
         insertBanner()
```
JoPintoPaul, line 63: I’m not sure how the original banner is triggered, but is there a risk that someone can have saved their cookie preferences, and the old banner is displayed and not removed? e.g. userPreference.getUserHasSavedCookiePreferences returns true but something triggers the banner.

JoPintoPaul (approved):

PR #82, opened 2020-10-08
Control optimizely via tracking.js

natcarey:


app/uk/gov/hmrc/trackingconsentfrontend/assets/src/interfaces/optimizelyCommunicatorFactory.ts
```diff
@@ -5,13 +5,37 @@ const optimizelyCommunicatorFactory = (window: Window): Communicator => ({
   sendPreferences: (userPreferences: UserPreferences) => {
 
     const setOptimizelyOptOut = (isOptOut: Boolean) => {
-      window.optimizely.push({ 'type': 'optOut', isOptOut });
+      window.optimizely.push({ 'type': 'optOut', isOptOut })
     }
 
-    window.optimizely = window.optimizely || [];
+    const setupOptimizelyIfConfigured = () => {
+      const regex = /^(\/|(https:)?\/\/www\.tax\.service\.gov\.uk\/|(https:)?\/\/www\.(qa|dev|staging)\.tax\.service\.gov\.uk\/|(http:)?\/\/localhost:12345\/)tracking-consent/
```
matthewmascord, line 12: This regex is something that is highly likely to need to change and require ongoing maintenance. For example, if the platform url structure changes, or the port number changes etc. The length of it also makes it difficult to easily see what it is doing. Additionally, I couldn't see any tests to exercise the different types of expected inputs. To make the code more flexible, I would probably aim to do something simpler like just doing a `.find` to look for the script tag containing the 'data-optimizely-id' attribute i.e. `scriptTags.find(scriptTag => scriptTag.getAttribute('data-optimizely-id') )`

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/interfaces/optimizelyCommunicatorFactory.ts
```diff
@@ -5,13 +5,37 @@ const optimizelyCommunicatorFactory = (window: Window): Communicator => ({
...
+      const isTruthy = x => !!x
+      const scriptTags = Array.from(document.getElementsByTagName('script'))
+      const trackingConsentScripts = scriptTags
+        .filter(scriptTag => (scriptTag.getAttribute('src') || '')
+          .match(regex))
+      const optimizelyIDs = trackingConsentScripts
```
matthewmascord, line 18: If we have already determined the current script, we don't need to generate an array of ids, we can just do a getAttribute on the current script. I'm not sure the isTruthy function is required and can be inlined.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/interfaces/optimizelyCommunicatorFactory.ts
```diff
@@ -5,13 +5,37 @@ const optimizelyCommunicatorFactory = (window: Window): Communicator => ({
...
+        .map(scriptTag => scriptTag.getAttribute('data-optimizely-id'))
+        .filter(isTruthy)
+      const firstTrackingConsentScriptTag = trackingConsentScripts[0]
+
+      optimizelyIDs.forEach(optimizelyId => {
+        if (!optimizelyId || !optimizelyId.match(/^\d{11}$/)) {
+          console.error('Optimizely ID is malformed', optimizelyId)
+        } else {
+          const script = document.createElement('script')
+          script.setAttribute('src', `https://cdn.optimizely.com/js/${optimizelyId}.js`)
```
matthewmascord, line 28: The base URL should be extracted as a constant e.g. in `constants/optimizely.ts`. Delegate url construction to a separate function.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/interfaces/optimizelyCommunicatorFactory.ts
```diff
@@ -5,13 +5,37 @@ const optimizelyCommunicatorFactory = (window: Window): Communicator => ({
...
+          firstTrackingConsentScriptTag.after(script)
+        }
+      })
+    }
+
+    window.optimizely = window.optimizely || []
     const preferences = userPreferences.getPreferences() || {}
     if (preferences.measurement === true) {
       setOptimizelyOptOut(false)
+      setupOptimizelyIfConfigured()
```
matthewmascord, line 38: This design currently implies that when the user accepts cookies, there is the potential for the content of the page to be rewritten and for them to see a different variation to the one they saw when they first visited the page.

app/uk/gov/hmrc/trackingconsentfrontend/assets/test/interfaces/optimizelyCommunicatorFactory.spec.ts
```diff
@@ -2,6 +2,43 @@ import optimizelyCommunicatorFactory from '../../src/interfaces/optimizelyCommun
 
 describe('sendPreferences', () => {
   let testScope
+
+  const getScripts = () => Array.from(document.getElementsByTagName('script'))
+  const getScriptSrcs = () => getScripts().map(e => e.getAttribute('src'))
+  const exampleTCScriptURL = '/tracking-consent/tracking.js'
+
+  const setupScriptWithID = (id, scriptSrc: string) => {
```
matthewmascord, line 10: I think this setup code can be simplified. We can just do something like:
```typescript
  const setupScriptWithID = (id, scriptSrc: string) => {
    document.getElementsByTagName('html')[0].innerHTML = ''
    const head = document.head
    const scriptTagWithAttr = document.createElement('script')
    if (scriptSrc) {
      scriptTagWithAttr.setAttribute('src', scriptSrc)
    }
    if (id) {
      scriptTagWithAttr.setAttribute('data-optimizely-id', id)
    }
    head.appendChild(scriptTagWithAttr)
  }
```
The first line resets the document so we don't need the cleanup.

app/uk/gov/hmrc/trackingconsentfrontend/assets/test/interfaces/optimizelyCommunicatorFactory.spec.ts
```diff
@@ -2,6 +2,43 @@ import optimizelyCommunicatorFactory from '../../src/interfaces/optimizelyCommun
...
+    if (scriptSrc) {
+      scriptTagWithAttr.setAttribute('src', scriptSrc)
+    }
+    if (id) {
+      scriptTagWithAttr.setAttribute('data-optimizely-id', id)
+    }
+    head.appendChild(scriptTagWithAttr)
+
+    return {
+      trigger: () => {
```
matthewmascord, line 23: I would just have trigger as a separate function as it doesn't depend on the consts above and perhaps rename to something like `sendMeasurementAllowed()`

app/uk/gov/hmrc/trackingconsentfrontend/assets/test/interfaces/optimizelyCommunicatorFactory.spec.ts
```diff
@@ -68,4 +106,85 @@ describe('sendPreferences', () => {
...
+      .toEqual([
+        exampleTCScriptURL,
+        'https://cdn.optimizely.com/js/98765432109.js'
+      ])
+    expect(console.error).not.toHaveBeenCalled()
+
+    control.cleanup()
+  })
+
+  it('should not setup optimizely when configured on the wrong script', () => {
```
matthewmascord, line 138: I would question whether we need to support this case and introduces the need to have that complicated regex or something similar. For me, it's sufficient to simply filter the script tags on data-optimizely-id.

app/uk/gov/hmrc/trackingconsentfrontend/assets/test/interfaces/optimizelyCommunicatorFactory.spec.ts
```diff
@@ -68,4 +106,85 @@ describe('sendPreferences', () => {
...
+    expect(getScriptSrcs())
+      .toEqual([
+        nonTrackingConsentScriptSrc
+      ])
+    expect(console.error).not.toHaveBeenCalled()
+
+    control.cleanup()
+  })
+
+  it('should not setup optimizely when configured on an inline script', () => {
```
matthewmascord, line 152: Again, questionable whether needed.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/interfaces/optimizelyCommunicatorFactory.ts
```diff
@@ -5,13 +5,37 @@ const optimizelyCommunicatorFactory = (window: Window): Communicator => ({
...
+        .filter(isTruthy)
+      const firstTrackingConsentScriptTag = trackingConsentScripts[0]
+
+      optimizelyIDs.forEach(optimizelyId => {
+        if (!optimizelyId || !optimizelyId.match(/^\d{11}$/)) {
+          console.error('Optimizely ID is malformed', optimizelyId)
+        } else {
+          const script = document.createElement('script')
+          script.setAttribute('src', `https://cdn.optimizely.com/js/${optimizelyId}.js`)
+          firstTrackingConsentScriptTag.after(script)
```
matthewmascord, line 29: `.after` is not supported in IE10. We would need to do something like:
```typescript
const scriptTagParentNode = scriptTag.parentNode
if (scriptTagParentNode !== null) {
  scriptTagParentNode.insertBefore(script, scriptTag.nextSibling)
}
```

matthewmascord (commented):
Thanks for taking the time to investigate this approach.

In addition to the inline comments:
* I think this needs some Selenium acceptance tests to prove that it works against the service test page.
* There is a technical risk that the solution will suffer from page flashing. I actually found this to be the case when testing locally. https://help.optimizely.com/Integrate_Other_Platforms/Implementing_Optimizely_with_a_tag_manager
* As it stands, the user will potentially be moved to a different variation of the page when they click 'Accept all cookies'.

Overall, I think this needs some further consideration of the technical risks and change of scope before I'd be happy with it being merged.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/interfaces/optimizelyCommunicatorFactory.ts
```diff
@@ -5,13 +5,37 @@ const optimizelyCommunicatorFactory = (window: Window): Communicator => ({
   sendPreferences: (userPreferences: UserPreferences) => {
 
     const setOptimizelyOptOut = (isOptOut: Boolean) => {
-      window.optimizely.push({ 'type': 'optOut', isOptOut });
+      window.optimizely.push({ 'type': 'optOut', isOptOut })
     }
 
-    window.optimizely = window.optimizely || [];
+    const setupOptimizelyIfConfigured = () => {
```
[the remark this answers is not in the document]
natcarey: That's fair, I've not refactored as it's a spike.

I'd like to discuss using this mechanism for any config including the GTM container reference.  I imagine we'd set up a central method like `getAttributeFromCurrentScript` if we're using this approach in multiple places.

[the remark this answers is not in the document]
matthewmascord: Agreed. Thinking about it, we actually might want to do something like:
```javascript
const scriptTag = document.querySelector('script[data-id="tracking-consent-frontend"]')
```
It's a shame document.currentScript does not exist in IE.

PR #85, opened 2020-10-09
PLATUI-649: Add auditing

matthewmascord:


app/uk/gov/hmrc/trackingconsentfrontend/assets/src/common/post.ts
```diff
@@ -0,0 +1,11 @@
+/* istanbul ignore file */
```
JoPintoPaul, line 1: What does this line mean?

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/config/config.ts
```diff
@@ -0,0 +1,21 @@
...
+        return undefined
+    }
+    const port: string | null = scriptTag.getAttribute('data-port')
+    if (port === null) {
+        return undefined
+    }
+    return port
+}
+
+export const getBaseUrl = (): string => {
```
JoPintoPaul, line 15: Might this function be better named `getLocalBaseUrl`?

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/config/config.ts
```diff
@@ -0,0 +1,21 @@
+export const getWelshLanguageEnabled = () => false
+
+export const getPort = (): string | undefined => {
```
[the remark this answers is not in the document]
matthewmascord: Good catch.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/domain/userPreferencesFactory.ts
```diff
@@ -50,15 +60,15 @@ const userPreferencesFactory = (): UserPreferences => {
...
 
-  const getPreferences = () => {
+  function getPreferences(): Preferences | undefined {
     const cookie = Cookies.getJSON(COOKIE_CONSENT)
     if (cookie === null || cookie === undefined || cookie.preferences === undefined) {
       return undefined
     }
 
-    const { version } = cookie
+    const { version, datetimeSet } = cookie
```
natcarey, line 71: It looks like this new variable is never used.  Can it be removed?

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/domain/userPreferencesFactory.ts
```diff
@@ -50,15 +60,15 @@ const userPreferencesFactory = (): UserPreferences => {
     return entries.filter(isPreferenceStrictlyBoolean)
   }
 
-  const allThePreferences = () => fromEntries(cookieTypes.map(cookieType => [cookieType, true]))
+  const allThePreferences = () => fromEntries(cookieTypes.map((cookieType) => [cookieType, true]))
 
-  const getPreferences = () => {
+  function getPreferences(): Preferences | undefined {
```
natcarey, line 65: The introduction of functions here instead of constants is a change in direction, what benefit do we get from that change?

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/config/config.ts
```diff
@@ -0,0 +1,21 @@
+export const getWelshLanguageEnabled = () => false
+
+const getPort = (): string | undefined => {
+    const scriptTag: HTMLScriptElement | null = document.querySelector('script[data-id="tracking-consent-frontend"]')
```
natcarey, line 4: I'm nervous about looking up the script tag in this way, it:

a) changes the interface with the services
b) allows any script tag to be wrongly identified as the tracking consent script tag

I think we should discuss the pros/cons of using the approach from my PR https://github.com/hmrc/tracking-consent-frontend/pull/82/files#diff-c50745f995c8fa68ee8fb65caa8781a4c7f5d5825a720d16dc76c357da88ab1aR12

I had previously discounted using the `currentScript` option, but actually it would keep more modern browsers safer so maybe we should use that and only fall back if it's not available.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/config/config.ts
```diff
@@ -0,0 +1,21 @@
...
+    if (scriptTag === null) {
+        return undefined
+    }
+    const port: string | null = scriptTag.getAttribute('data-port')
```
natcarey, line 8: I'm nervous about reading the port from a data attribute which could accidentally or maliciously be incorrect.  We know the port already, or we can read it from the `src` attribute.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/interfaces/getMessages.ts
```diff
@@ -5,7 +5,7 @@ import messagesCy from '../../../../../../../../conf/messages.cy';
 
 import getLanguage from "./getLanguage";
 import {CY} from "../constants/languages";
-import getWelshLanguageEnabled from "./getWelshLanguageEnabled";
+import { getWelshLanguageEnabled } from "../config/config";
```
natcarey, line 8: `config/config` doesn't feel like a great path - can we use `./config` or something more specific instead of the second `config`?

app/uk/gov/hmrc/trackingconsentfrontend/assets/test/common/getTrackingConsentBaseUrl.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+    }
+    const url = getTrackingConsentBaseUrl()
+
+    expect(url).toEqual('')
+  })
+
+  it('should return "http://localhost:12345" if running under http://localhost:9000/some-service', () => {
+    // @ts-ignore
+    window.location = {
+      host: 'localhost:12345'
```
natcarey, line 24: This doesn't match the test name (http://localhost:9000/some-service)

app/uk/gov/hmrc/trackingconsentfrontend/assets/test/config/config.spec.ts
```diff
@@ -0,0 +1,72 @@
+import '@testing-library/jest-dom/extend-expect'
+import { getLocalBaseUrl } from '../../src/config/config'
+
+describe('config', () => {
+  beforeEach(() => {
+    document.getElementsByTagName('html')[0].innerHTML = `<html>
+    <head>
+        <script src="/another-script.js"></script>
+        <script data-id="tracking-consent-frontend" data-port="54321" src="/bundle.js"></script>
```
natcarey, line 9: In what scenario would it be desirable to point to a port other than the one tracking consent is being loaded from?

natcarey (changes requested):

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/config/config.ts
matthewmascord: Why? What threat do you have in mind? How would an attacker be able to maliciously alter it?

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/domain/userPreferencesFactory.ts
matthewmascord: Ability to use `this` which is more idiomatic Javascript.

app/uk/gov/hmrc/trackingconsentfrontend/assets/test/config/config.spec.ts
matthewmascord: It is needed for acceptance testing. In Jenkins, the tracking consent service runs under a different port.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/config/config.ts
matthewmascord: currentScript is not available in IE.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/domain/userPreferencesFactory.ts
matthewmascord: Good catch, removed.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/config/config.ts
matthewmascord: Port now extracted from src of current script.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/interfaces/getMessages.ts
matthewmascord: I've moved these functions into /common for now.

app/uk/gov/hmrc/trackingconsentfrontend/assets/test/common/getTrackingConsentBaseUrl.spec.ts
matthewmascord: Good catch, amended.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/config/config.ts
matthewmascord: In the meeting on 19 October 2020, we agreed the principle of using an identifier to look up the current script. We also talked about using getElementById. However, on reflection that may be a little more risky than above due to the fact that this will pollute the id namespace and potentially cause problems if any other elements are defined with that id, unless we use very unique id like a GUID. The code above will only cause problems if another script tag has been defined with data-id="tracking-consent-frontend", which seems unlikely in my opinion.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/common/getCurrentScript.ts
```diff
@@ -0,0 +1,9 @@
+const getCurrentScript = (): HTMLScriptElement | undefined => {
+    const scriptTag: HTMLScriptElement | null = document.querySelector('script[data-id="tracking-consent-frontend"]')
```
natcarey, line 2: Can we change this to `id` instead of `data-id` please?

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/common/getLocalBaseUrl.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    const src: string | null = scriptTag.getAttribute('src')
+    if (src === null) {
+        return undefined
+    }
+
+    return extractPortFromUrl(src)
+}
+
+const getLocalBaseUrl = (): string => {
```
natcarey, line 28: Can you consolidate the full Base URL logic in here?  This does half the job and `getTrackingConsentBaseUrl` does the other half, I'd prefer to see that all in together.

natcarey: Looking at this more closely I'd suggest this:

```
const extractPortFromUrl = (src) => {
    var anchor = document.createElement("a");
    anchor.href = src;
    const host = anchor.host
    const protocol = anchor.protocol
    return [protocol, host].join('//')
}
```

I've tried that in the console feeding in inputs like:

`extractPortFromUrl('https://github.com/abc')`
`extractPortFromUrl('https://github.com:1234/abc')`
`extractPortFromUrl('/abc')`

I think this method is all you need to turn any URL into a baseUrl.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/domain/userPreferencesFactory.ts
natcarey: I personally really dislike using `this` in JavaScript for stuff like (no pun intended) this.  It is more efficient but can cause problems if methods are called out of context - I find the simplicity of being able to do things like `eventEmitter.on('abc', userPreferences.userAcceptsAll)` is worth the compromise.  We can get around it with `bind`, `call` and `apply` but I don't think we gain enough to justify the inconvenience.

test/acceptance/specs/AuditingSpec.scala
```diff
@@ -0,0 +1,113 @@
...
+    click on useMarketingCookiesLabel
+    click on useSettingsCookiesLabel
+
+    And("clicks submit")
+    endpointServer.resetRequests()
+    click on submitButton
+
+    Then("an audit event is sent")
+    eventually {
+      val requests = endpointServer.findAll(anyRequestedFor(anyUrl()))
```
natcarey, line 62: Do we have tests to show the URL is correct?

app/uk/gov/hmrc/trackingconsentfrontend/assets/test/common/getLocalBaseUrl.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+          <script src="/another-script.js"></script>
+      </head>
+      <body>
+      </body>
+  </html> 
+  `
+
+    const baseUrl = getLocalBaseUrl()
+
+    expect(baseUrl).toEqual('')
```
natcarey, line 84: If you take on my suggestion above some of these outputs will change.

Should we be testing for what happens when no script tag is found?  Should we be testing for when multiple script tags are found?

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/domain/userPreferencesFactory.ts
matthewmascord: Looking at this again there's actually a more subtle reason I changed it to use this. I ran it through a strict linter and saw that 'self' was being used before it was defined in line 20.

test/acceptance/specs/AuditingSpec.scala
matthewmascord: The stub is only set up for the correct URL so we would get a failure if attempting to POST to the wrong URL.

app/uk/gov/hmrc/trackingconsentfrontend/assets/test/common/getLocalBaseUrl.spec.ts
matthewmascord: Having moved this logic into getTrackingConsentBaseUrl, there are indeed tests for this situation.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/common/getLocalBaseUrl.ts
matthewmascord: The existing logic first tests if you are running locally, if not it returns a *relative* URL. It's only if you are running locally, it then performs the logic to extract the port and construct a localhost URL. This means in the normal case of running in production, the logic is simplest and doesn't involve parsing any URLs, which feels more robust. If we are running locally then it goes to the trouble of parsing the URL, but then we know we are running locally so don't need the full base URL.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/domain/userPreferencesFactory.ts
```diff
@@ -2,46 +2,56 @@
...
   }
 
-  const storePreferences = (preferences) => {
-    Cookies.set(COOKIE_CONSENT, {
-      version: '2020.1',
-      datetimeSet: new Date().toISOString(),
-      preferences: preferences
-    }, { sameSite: 'strict', expires: 3650 })
+  function sendPreferences(event: string) {
+    subscribers.forEach((subscriber) => subscriber.sendPreferences(this, event))
```
matthewmascord, line 20: Previously this used 'self' which is being used before it had been defined.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/domain/userPreferencesFactory.ts
natcarey: One good option is:

```
const self = {}

self.abc = () => {}

self.def = () => {
  self.abc()
}

return self
```

matthewmascord: I'm not keen on mutating self if we can avoid it. What about converting to a standard ES6 class?

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/common/getLocalBaseUrl.ts
natcarey: I'd prefer a simpler function with fewer lines which doesn't have forked logic for local environments.  Can you see a scenario which isn't covered by that approach?

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/domain/userPreferencesFactory.ts
matthewmascord: Removed use of `this` in latest commit. Resolved ordering issue by declaring self at top and using functions for methods that are exported. This works because functions are hoisted. I feel this is a good compromise.

JoPintoPaul:
Closing as this has been superseded by PR #88

PR #90, opened 2020-11-02
Platui 811 mandate new containers

JoPintoPaul:


JoPintoPaul:
As part of this, I have at the moment put in console warn in the `pageHandler` if the gtm containerId is undefined, leaving the error to be thrown when we try to ENABLE gtm. However, happy to discuss this.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/common/pageHandler.ts
```diff
@@ -5,13 +5,13 @@ import { enableTrackingConsent } from "../constants/featureNames";
 import { SERVICE_PAGE_LOAD_EVENT } from "../constants/events";
 import getGtmContainerId from "./getGtmContainerId";
 
-const pageHandler = (document: HTMLDocument, userPreference: UserPreferences, pageRenderer, containerId) => {
+const pageHandler = (document: HTMLDocument, userPreference: UserPreferences, pageRenderer) => {
     const gtmContainerId = getGtmContainerId()
     if (gtmContainerId === undefined) {
-        console.warn("Deprecation notice: container id not specified in data-gtm-container attribute. The functionality you are currently using will be removed on 03 November 2020")
+        console.error("container id not specified in data-gtm-container attribute. The previous default container ID was removed on 03 November 2020")
```
natcarey, line 11: I think we should be stopping execution here - maybe a `throw new Error` or maybe `return`/`else` after the console error.

JoPintoPaul: @natcarey That seems fair enough - would you also keep in the thrown error within `enableGtm` if the container id is undefined?

JoPintoPaul: @natcarey If we error here, then this would cause an error even before the feature is enable - is that definitely what we want?

natcarey: Hmm, I guess the actual error here is that GTM can't be included, so maybe it's just about putting `enableGtm(gtmContainerId)` into an `else` block would be the best representation of that.  The logic there is "if you provide a GTM container reference we'll include GTM".

matthewmascord: I would go with throwing in pageHandler as Nat suggests because it doesn't make sense to continue without gtm.

matthewmascord (approved):

PR #96, opened 2020-11-18
PLATUI-828: add maintainers documentation

matthewmascord:


natcarey (approved):

docs/adr/0003-use-markdown-architectural-decision-records.md
```diff
@@ -0,0 +1,61 @@
+# Use Markdown Architectural Decision Records (MADRs)
```
JoPintoPaul, line 1: Should this be 0001?

PR #97, opened 2020-11-20
PLATUI-849: Fix a11y and ZAP tests

matthewmascord:


test/resources/application.conf
```diff
@@ -64,6 +64,9 @@ zap-automation-config {
   ]
 
   alertsToIgnore = [
+    # enableTrackingConsent is a non-HttpOnly cookie we set and need to be readable in Javascript
```
JoPintoPaul, line 67: Shall we add to this comment that it is a temporary feature and flag and will be deprecated by the end of 2020?

JoPintoPaul (approved):

PR #100, opened 2020-12-03
Added eslint, fixed the issues it had.

natcarey:


app/uk/gov/hmrc/trackingconsentfrontend/assets/src/common/createNonce.ts
```diff
@@ -1,7 +1,8 @@
-import getNonce from "./getNonce";
+import getNonce from './getNonce';
 
 // The following is required for style-loader to add the nonce attribute
 // correctly to the cookie banner style element
 // It must appear exactly as below (no var, let or const)
 // @ts-ignore
-__webpack_nonce__ = getNonce()
+// eslint-disable-next-line no-underscore-dangle
+window.__webpack_nonce__ = getNonce();
```
matthewmascord, line 8: Has this change been tested? When I implemented it it had to be exactly as it was before for webpack to pick it up, hence the @ts-ignore.

natcarey: Fair question.  I haven't, let me see if I can ignore that line instead.

natcarey: Updated to avoid changing the production code.

docs/adr/0006-use-eslint-for-linting-and-formatting.md
```diff
@@ -0,0 +1,60 @@
...
+* Files within a project will be consistent with each other
+* When widely applied this Javascript/Typescript files between projects will be consistent
+* IDE will reformat for us which takes away the need for us to do it ourselves
+* Code reviews will focus less on code format and more on the details of the changes made
+* Builds will fail if format isn't followed which guarentees consistency
+
+### Negative Consequences
+
+* Builds will fail if format isn't followed which can disrupt focus/workflow (mitigated by pre-commit hooks)
+cd
```
matthewmascord, line 60: Typo?

app/uk/gov/hmrc/trackingconsentfrontend/assets/.eslintrc.json
```diff
@@ -0,0 +1,47 @@
+{
+  "extends": [
+    "airbnb-base"
```
matthewmascord, line 3: Did you try using eslint-config-airbnb-typescript: https://www.npmjs.com/package/eslint-config-airbnb-typescript? It might simplify this configuration.

app/uk/gov/hmrc/trackingconsentfrontend/assets/package.json
```diff
@@ -59,9 +59,15 @@
...
     "@types/jest": "25.1.4",
-    "@typescript-eslint/eslint-plugin": "2.23.0",
-    "@typescript-eslint/parser": "2.23.0",
+    "@typescript-eslint/eslint-plugin": "^2.23.0",
+    "@typescript-eslint/parser": "^2.23.0",
     "backstopjs": "5.0.4",
+    "eslint": "^7.2.0",
+    "eslint-config-airbnb": "^18.2.1",
+    "eslint-plugin-import": "^2.22.1",
+    "eslint-plugin-jsx-a11y": "^6.4.1",
```
matthewmascord, line 68: I don't think we need these jsx and react specific dependencies.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/domain/userPreferencesFactory.ts
```diff
@@ -1,99 +1,99 @@
...
-    sendPreferences,
-    subscribe
-  }
-
-  return self
-}
-
-export default userPreferencesFactory
+    subscribe,
+    userAcceptsAll: () => {
```
matthewmascord, line 73: For me having all of the functions inline like this now does make it harder to read. Is it possible to keep eslint happy while retaining the previous structure? I know we've been over this before but this feels like a backwards step.

app/uk/gov/hmrc/trackingconsentfrontend/assets/.eslintrc.json
natcarey: I'll take a look.

app/uk/gov/hmrc/trackingconsentfrontend/assets/package.json
natcarey: Yeah, I don't think we need that.  I'll remove it and confirm.

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/domain/userPreferencesFactory.ts
natcarey: The problem here is that it's particular about the order lines of code appear - the code was never a problem here but it was a challenge to persuade eslint that it wasn't a problem :)

If you'd like to suggest an alternative I'd be happy to check it out.

matthewmascord: I'm not sure you liked my previous suggestions :-) Should we just ignore this rule then with a comment?

app/uk/gov/hmrc/trackingconsentfrontend/assets/.eslintrc.json
natcarey: I've made an update which now uses that.

app/uk/gov/hmrc/trackingconsentfrontend/assets/package.json
natcarey: There are two different options in the documentation https://www.npmjs.com/package/eslint-config-airbnb-typescript

The "If you don't need React support" section doesn't actually work so I used the react section which does work.  It would be good to fix this because I have also had to add this which I'd rather remove:

```
      "settings": {
        "react": {
          "version": "latest"
        }
      }
```

app/uk/gov/hmrc/trackingconsentfrontend/assets/src/domain/userPreferencesFactory.ts
natcarey: I'm happy to do that.

app/uk/gov/hmrc/trackingconsentfrontend/assets/.eslintrc.json
matthewmascord: Wouldn't you need to do ```extends: ['airbnb-typescript'],``` in your .eslintrc.json?

app/uk/gov/hmrc/trackingconsentfrontend/assets/.eslintrc.json
```diff
@@ -7,40 +7,34 @@
...
   "overrides": [
     {
       "files": [
         "*.ts"
       ],
-      "parser": "@typescript-eslint/parser",
-      "parserOptions": {},
-      "plugins": [
-        "@typescript-eslint"
+      "extends": [
```
matthewmascord, line 19: Shouldn't this go in line 2?

natcarey: Updated, thanks.

matthewmascord (approved):

PR #113, opened 2021-03-02
PLATUI-550: Link to cookie-details page in help-frontend

matthewmascord:


conf/application.conf
```diff
@@ -151,3 +151,8 @@ play.filters.cors {
   allowedHttpMethods = ["POST", "OPTIONS"]
   allowedOrigins = ["https://www.tax.service.gov.uk"]
 }
+
+help-frontend {
+  host = "http://localhost:9240"
```
JoPintoPaul, line 156: Should this be added to the `app-config-env` files?

conf/messages.en
```diff
@@ -11,7 +11,7 @@ banner.paragraph1.afterLink = about how you use HMRC services. We use this infor
 banner.acceptAllButton = Accept all cookies
 banner.setCookiePreferencesButton = Set cookie preferences
 
-banner.saveConfirmation.paragraph1.beforeLink = You’ve accepted all cookies. You can 
+banner.saveConfirmation.paragraph1.beforeLink = You’ve accepted all cookies. You can
```
JoPintoPaul, line 14: Are these end of line spaces needed?

conf/application.conf
matthewmascord: On an environment it will not use this config, this is needed for local development only.

conf/messages.en
matthewmascord: There is whitespace in the template, so I don't think so.

JoPintoPaul: That's cool, in `accessibility-statement-frontend` I think we have a convention of having whitespace within messages rather than template, but also we use whitespace in templates for markup so it's not something I'm too bothered about.

JoPintoPaul (approved):

conf/messages.en
matthewmascord: That convention obviously passed me by :-)

PR #123, opened 2021-03-15
PLATUI-1065: Send GTM events/variables before GTM is enabled, only se…

matthewmascord:
…nd Optimizely opt outs when banner is enabled


JoPintoPaul:
@matthewmascord Everything here looks good to me. Should we add a test for the Optimizely cookie being set at the correct time in the acceptance tests? At the moment, we test functionality in isolation but an integrated test seems useful.
https://github.com/hmrc/tracking-consent-frontend/blob/master/test/acceptance/specs/ServiceTestPageSpec.scala

JoPintoPaul (approved):

PR #131, opened 2021-03-26
Update alert-filters.json

rewtd:


vivrichards600 (approved):
Double checked ruleId against lookup https://github.com/zaproxy/zaproxy/blob/main/docs/scanners.md - all looks good :)

rewtd:
I do not have write access but if this should get merged will correct the report.

matthewmascord (approved):

PR #138, opened 2021-06-03
PLATUI-1214: Added missing data-language attribute in SettingsPageScr…

JoPintoPaul:
…ipts


app/uk/gov/hmrc/trackingconsentfrontend/assets/src/ui/renderSettingsSaveConfirmationMessage.ts
```diff
@@ -8,12 +8,13 @@ import callIfNotNull from '../common/callIfNotNull';
 import scrollToTop from '../common/scrollToTop';
 import getReferrer from '../common/getReferrer';
 // @ts-ignore
```
natcarey, line 10: Do we still need the @ts-ignore?

matthewmascord: Indeed, should be able to remove that now.

matthewmascord (approved):

PR #139, opened 2021-06-04
PLATUI-1189: move javascript assets to top level for dependabot alerts

matthewmascord:
Also uplifts some dev dependencies to fix vulnerabilities.


oscarduignan (approved):
looks good to me

PR #143, opened 2021-06-24
Plat UI 1250 fix vulnerabilities

natcarey:
Most issues are solved by an automated fix (first commit), some required manual updates (second commit).

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,8 +36,6 @@
     ]
   },
   "dependencies": {
-    "govuk-frontend": "3.11.0",
-    "js-cookie": "2.2.1",
     "@babel/core": "7.10.1",
     "@babel/plugin-transform-runtime": "7.10.1",
     "@babel/preset-env": "7.10.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -36,22 +36,22 @@
...
     "@babel/preset-env": "7.10.1",
     "babel-loader": "8.1.0",
     "core-js": "3.6.5",
     "css-loader": "5.2.6",
     "dot-properties": "1.0.1",
-    "node-sass": "4.14.1",
-    "sass-loader": "8.0.2",
+    "govuk-frontend": "3.11.0",
+    "js-cookie": "2.2.1",
+    "node-sass": "^6.0.1",
```
matthewmascord, line 48: In our chat with platform security, I think we were advised to pin all dependencies (so not use ^)

natcarey: Cool.  I'll do that.  Just for the ones I'm changing or shall I do them all?

matthewmascord: If there were any not pinned before then probably makes sense to do for all.

matthewmascord (approved):

PR #147, opened 2021-07-19
Using REM font sizes

natcarey:
I also manually coppied backstop tests from Jenkins as my local showed different results.


natcarey:
It's worth noting that the local VRT passes when the baseline is refreshed before [The REM flag](https://github.com/hmrc/tracking-consent-frontend/pull/147/files#diff-69e6d789980d76e639adda967c3c0ee399443c46af55cd9f9b5902edc1904d87R4) was changed.  The display is different in Jenkins, the font appears smaller but this could not be reproduced in my local system.

oscarduignan:
@natcarey I think it might be worth working out what's causing this difference on jenkins before we merge. If it's not working on jenkins maybe there are other services this won't work on. The smaller for everyone text to me seems like it might be a bigger accessibility problem than having text that doesn't increase with base font size

https://github.com/hmrc/tracking-consent-frontend/blob/00bcbf22c7eebaf1f6e9bcb6b0ffb6dd23208ad1/app/uk/gov/hmrc/trackingconsentfrontend/assets/test/backstop/backstop_data/bitmaps_reference/backstop_default_Tracking_Consent_-_Cookie_Banner_-_Assets_Frontend_-_Click_accept_additional_0_document_0_phone.png

https://github.com/hmrc/tracking-consent-frontend/blob/107dddcbb4cb0b8e32ffb37776c338d842e0cf0b/js/test/backstop/backstop_data/bitmaps_reference/backstop_default_Tracking_Consent_-_Cookie_Banner_-_Assets_Frontend_-_Click_accept_additional_0_document_0_phone.png

matthewmascord:
As mentioned, assets-frontend does not support our using rem because it overrides the root font size ([here](https://github.com/hmrc/assets-frontend/blob/6e3655d7f10909d7c6a9e7460c8f3b17684cb9e7/template/assets/stylesheets/govuk-template.css#L183)) If you take a look at the 'rich diff' above, you'll see the new reference images have the font size too small in the banner.

Re WCAG, supporting the ability to increase font size to 200% using zoom is sufficient for a pass under WCAG criterion 1.4.4 (Resize Text).

matthewmascord:
Closing because this is not achievable while supporting assets-frontend. The above changes break the rendering of font-sizes.

PR #149, opened 2021-08-18
PLATUI-1332: Uplift to v1.0.0 of play-frontend-hmrc and fix npm audit…

matthewmascord:
… failures

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11496,9 +11496,9 @@
       "integrity": "sha1-QRyttXTFoUDTpLGRDUDYDMn0C0A="
     },
     "path-parse": {
-      "version": "1.0.6",
-      "resolved": "https://registry.npmjs.org/path-parse/-/path-parse-1.0.6.tgz",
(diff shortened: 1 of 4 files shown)


package-lock.json
```diff
@@ -11496,9 +11496,9 @@
       "integrity": "sha1-QRyttXTFoUDTpLGRDUDYDMn0C0A="
     },
     "path-parse": {
-      "version": "1.0.6",
-      "resolved": "https://registry.npmjs.org/path-parse/-/path-parse-1.0.6.tgz",
-      "integrity": "sha512-GSmOT2EbHrINBf9SR7CDELwlJ8AENk3Qn7OikK4nFYAu3Ote2+JYNVvkpAEQm3/TLNEJFD/xZJjzyxg3KBWOzw=="
+      "version": "1.0.7",
```
oscarduignan, line 11499: @matthewmascord all our versions in package.json are fixed - is there a missing change to package.json that bumps one of our versions there? may well have misunderstood part of how npm audit fixes stuff but wanted to check

oscarduignan (approved):

package-lock.json
matthewmascord: As this is a transitive dependency, I'm not sure there's much we can do in package.json to fix this issue. package-lock.json is ultimately the source of truth for the versions used, so everything is pinned there. Having the versions pinned in package.json is a further line of defence in case the package-lock.json is deleted unknowingly.

PR #159, opened 2021-11-09
Update npm deps using npm audit fix

oscarduignan:
this is a work in progress, ran out of time yesterday


oscarduignan:
We had to raise the PR, let jenkins run the VRT tests, then copy the screenshots from jenkins down to our local version and force push them up again because locally VRT renders text a bit differently.

PR #204, opened 2022-11-07
NOJIRA: Updated dependencies for npm vulnerability in loader-utils

JoPintoPaul:


oscarduignan (approved):
looks good to me, might be worth doing a quick check in IE11 to in staging when deployed to make sure it works in case ts / babel bits dropped support for anything

PR #213, opened 2022-12-09
PLATUI-2109 retry flickering tests; fix some given/when/then oddities

ellamdav:
From a bit of searching, it seems that the 502 Bad Gateway error may indicate a cache problem with the browser.  Rather than go down a rabbit hole trying to diagnose and fix that, I've added code to retry any tests that fail and have the `Retryable` tag.

### DIFF
--- a/test/acceptance/specs/CookieSettingsPageSpec.scala
+++ b/test/acceptance/specs/CookieSettingsPageSpec.scala
@@ -19,6 +19,7 @@ package acceptance.specs
 import acceptance.pages.CookieSettingsPage
 import acceptance.pages.CookieSettingsPage._
 import acceptance.specs.tags.Local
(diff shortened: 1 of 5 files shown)


test/acceptance/specs/CookieSettingsPageSpec.scala
```diff
@@ -115,11 +116,6 @@ class CookieSettingsPageSpec extends BaseAcceptanceSpec {
 
     And("the dataLayer contains the 'trackingConsentSettingsAccepted' event")
     settingsAllowedGtmEvent should not be null
-
```
ellamdav, line 118: this is outside the scope of this test of GTM events - there's a specific test for setting the cookie (the next one)

kyle-bowden (approved):

PR #220, opened 2023-01-13
NOJIRA npm audit - update npm audit dependencies

ellamdav:


oscarduignan:
Still go back and forth if we should bump the package.json version too since some deps have changed 🤷‍♂️

oscarduignan (approved):

PR #221, opened 2023-01-27
PLATUI-2190: update README

oscarduignan:
Describe what happens when a user revokes previously granted tracking consent. We got a question asking if the cookies needed to be removed and Holly confirmed with Ori that they don't need to be removed to comply with PECR rules.

### DIFF
--- a/app/uk/gov/hmrc/trackingconsentfrontend/controllers/AuditController.scala
+++ b/app/uk/gov/hmrc/trackingconsentfrontend/controllers/AuditController.scala
@@ -23,5 +23,5 @@ import scala.concurrent.Future
 
 @Singleton
 class AuditController @Inject() (mcc: MessagesControllerComponents) extends FrontendController(mcc) {
-  def audit: Action[AnyContent] = Action.async(implicit request => Future.successful(Ok))
+  def audit: Action[AnyContent] = Action.async(Future.successful(Ok))
 }
--- a/app/uk/gov/hmrc/trackingconsentfrontend/testonly/views/TestPage.scala.html
+++ b/app/uk/gov/hmrc/trackingconsentfrontend/testonly/views/TestPage.scala.html
@@ -22,7 +22,7 @@
 )
 @()(implicit request: Request[_], messages: Messages, appConfig: AppConfig)
 
-@defining(Messages("testPage.title"), Messages("testPage.heading")) { case (title, heading) =>
+@defining((Messages("testPage.title"), Messages("testPage.heading"))) { case (title, heading) =>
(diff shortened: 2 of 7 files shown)


app/uk/gov/hmrc/trackingconsentfrontend/controllers/AuditController.scala
```diff
@@ -23,5 +23,5 @@ import scala.concurrent.Future
 
 @Singleton
 class AuditController @Inject() (mcc: MessagesControllerComponents) extends FrontendController(mcc) {
-  def audit: Action[AnyContent] = Action.async(implicit request => Future.successful(Ok))
+  def audit: Action[AnyContent] = Action.async(Future.successful(Ok))
```
oscarduignan, line 26: does it need to be async? 🤷 just removed the unused request because of a warning since I removed the silencer plugin because there wasn't a version of it for the new scala version

app/uk/gov/hmrc/trackingconsentfrontend/testonly/views/TestPage.scala.html
```diff
@@ -22,7 +22,7 @@
 )
 @()(implicit request: Request[_], messages: Messages, appConfig: AppConfig)
 
-@defining(Messages("testPage.title"), Messages("testPage.heading")) { case (title, heading) =>
+@defining((Messages("testPage.title"), Messages("testPage.heading"))) { case (title, heading) =>
```
oscarduignan, line 25: it did work before, but this is an example of where I think before we silenced all warnings from templates and now we're just silencing unused imports so this surfaced - it wanted an explicit tuple

app/uk/gov/hmrc/trackingconsentfrontend/views/components/Layout.scala.html
```diff
@@ -19,7 +19,7 @@
 @(pageTitle: Option[String] = None,
         headBlock: Option[Html] = None,
         scriptsBlock: Option[Html] = None
-)(contentBlock: Html)(implicit request: Request[_], messages: Messages, appConfig: AppConfig)
+)(contentBlock: Html)(implicit request: Request[_], messages: Messages)
```
oscarduignan, line 22: not used anymore since it's injected into hmrcLayout

test/unit/testonly/views/TestPageSpec.scala
```diff
@@ -16,7 +16,6 @@
 
 package unit.testonly.views
 
-import play.api.test.Helpers._
```
oscarduignan, line 19: was unused

ellamdav (approved):

JoPintoPaul (approved):

PR #223, opened 2023-02-20
PLATUI-2211: Added no print class to banner and uplifted library depe…

JoPintoPaul:
…ndencies

### DIFF
--- a/js/src/constants/cssClasses.ts
+++ b/js/src/constants/cssClasses.ts
@@ -15,5 +15,6 @@ export const GOV_UK_BODY_CLASS = 'govuk-body';
 export const GOV_UK_NOTIFICATION_LINK_CLASS = 'govuk-notification-banner__link';
 export const GOV_UK_SKIP_LINK_CLASS = 'govuk-skip-link';
 export const SKIP_LINK_CONTAINER_ID = 'skiplink-container';
+export const GOV_UK_NONE_PRINT_CLASS = 'govuk-!-display-none-print';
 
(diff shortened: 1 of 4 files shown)


js/src/constants/cssClasses.ts
```diff
@@ -15,5 +15,6 @@ export const GOV_UK_BODY_CLASS = 'govuk-body';
 export const GOV_UK_NOTIFICATION_LINK_CLASS = 'govuk-notification-banner__link';
 export const GOV_UK_SKIP_LINK_CLASS = 'govuk-skip-link';
 export const SKIP_LINK_CONTAINER_ID = 'skiplink-container';
+export const GOV_UK_DISPLAY_PRINT_NONE = 'govuk-!-display-none-print';
```
ellamdav, line 18: should this be `GOV_UK_DISPLAY_NONE_PRINT` for consistency? (although it feels like their class probably should have been `govuk-!-display-non-print` to start with 🤔 ).  or maybe follow the skip-link convention and use something like `GOV_UK_NON_PRINT_CLASS` ?

ellamdav (approved):

js/src/constants/cssClasses.ts
JoPintoPaul: @ellamdav Good spot - that's actually just a mis-transliteration from me, so I'll fix it to `GOV_UK_NONE_PRINT_CLASS` for consistency and clarity.

ellamdav (approved):

PR #231, opened 2023-09-18
NOJIRA suppress failing tests

KatiePocklington:


test/acceptance/specs/ServiceTestPageSpec.scala
```diff
@@ -140,15 +140,17 @@ class ServiceTestPageSpec extends BaseAcceptanceSpec {
     }
 
     Scenario("The user visits a page and the cookie banner is displayed") {
-      Given("the user clears their cookies")
-      deleteAllCookies()
+      pendingUntilFixed {
```
JoPintoPaul, line 143: Almost right, just this wrapper is around the wrong test - I'll add in the correct place as a suggestion.

JoPintoPaul (approved):

PR #254, opened 2024-02-29
Update readme for running acceptance tests

kyle-bowden:


JoPintoPaul:
@kyle-bowden Looking at the `run.sh` script, I wonder if this would be good to rename whilst we're making changes? Something like `run-with-test-endpoints.sh`?

PR #260, opened 2024-04-18
PLATUI-2883 resync bundled cookie banner styles

oscarduignan:
just creating draft for some comments


js/styles/_typography.scss
```diff
@@ -13,10 +41,148 @@ $govuk-typography-use-rem: false;
...
+
+  margin-top: 0;
+  @include govuk-responsive-margin(6, "bottom");
+}
+
+.govuk-body-l {
+  @extend %govuk-body-l;
+}
+
 %govuk-body-m {
```
oscarduignan, line 109: I think before we just had the % because it doesn't result in any output - so a class selector like `.govuk-body-m` will result in some css output, however a placeholder selector `%govuk-body-m` will be available to be extended by other stuff but by itself not result in any output. I guess we define it in here, maybe because govuk has the classes and placeholders in the same class so by importing one we got extra classes being output. Where we just wanted to be able to:
```scss
.cbanner-govuk-heading-m {
  @extend %govuk-heading-m;
}
```
in govuk-frontend.scss without impacting other styles on the page / having more css to inject that we need by just @import'ing it inside a class to avoid pollution

oscarduignan: I kind of wonder if it's still the case we have to do this or if sass in govuk-frontend has been refactored a bit - also seems like it would be a good idea to have some tests that say "we don't redefine styles for govuk classes" so do a grep on the output css to ensure there is no `.govuk-*` in there

js/styles/_button.scss
```diff
@@ -219,29 +204,39 @@ $button-shadow-size: $govuk-border-width-form-element;
...
-      background-color: $govuk-warning-button-colour;
+      background-color: $govuk-inverse-button-colour;
     }
   }
 }
 
 .cbanner-govuk-button--start {
   @include govuk-typography-weight-bold;
-  @include govuk-typography-responsive($size: 24, $override-line-height: 1);
+  @include govuk-font-size($size: 24, $line-height: 1);
```
oscarduignan, line 239: @ellamdav it seems like govuk-font-size doesn't set the font-family https://github.com/alphagov/govuk-frontend/blob/1871898df79e4c849f89d58f7bb10fb400aed436/packages/govuk-frontend/src/govuk/helpers/_typography.scss#L160-L237 guess they made it more specialised - I think we need just `govuk-font` https://github.com/alphagov/govuk-frontend/blob/1871898df79e4c849f89d58f7bb10fb400aed436/packages/govuk-frontend/src/govuk/helpers/_typography.scss#L253

oscarduignan: oh nevermind, spotted govuk-font is used further up 🤦‍♂️

js/styles/_typography.scss
ellamdav: Is there anything you want to change / add now, related to the above comments?

js/styles/_typography.scss
```diff
@@ -1,7 +1,35 @@
 // Mixins taken from node_modules/govuk-frontend/govuk/core/_typography
-@import "node_modules/govuk-frontend/govuk/base";
+@import "node_modules/govuk-frontend/dist/govuk/base";
 
-$govuk-typography-use-rem: false;
```
oscarduignan, line 4: we should remove all the extra styles in here, it should just be % prefixed bits- and only the heading-m and body-m

if we don't we'll end up changing styles on the page (possibly)

js/styles/_typography.scss
oscarduignan: we should remove all the extra styles in here, it should just be % prefixed bits- and only the heading-m and body-m

if we don't we'll end up changing styles on the page (possibly)

oscarduignan: we could also remove the
```
@import "node_modules/govuk-frontend/dist/govuk/base";
```
because we import it in the govuk-frontend.scss above this file

PR #261, opened 2024-04-18
PLATUI-2883: wait for fonts

oscarduignan:


oscarduignan:
haven't recreated the issue, but doesn't appear to work unfortunately

PR #262, opened 2024-04-18
PLATUI-2883: config backstop with more precise matching

oscarduignan:
the wait-for-fonts this is based on might not work (didn't actually recreate the issue), so don't merge this to main


oscarduignan:
sorted in another PR

PR #274, opened 2024-05-30
PLATUI-3063 update js-cookie and refactor

TimothyFothergill:
# Purpose of PR
To bump the version of js-cookie, the update is to a new major version of the library, removing a few methods:

- `getJSON()` is now `Cookies.get('foo')`, wrapped in a `JSON.parse()`.
- `set()` now doesn't automatically stringify JSON.

### DIFF
--- a/js/test/domain/userPreferencesFactory.spec.ts
+++ b/js/test/domain/userPreferencesFactory.spec.ts
@@ -17,7 +17,7 @@ describe('userPreferencesFactory', () => {
     spyOn(Cookies, 'set').and.callFake((cookieName, value) => {
       testScope.cookieData[cookieName] = value;
     });
-    spyOn(Cookies, 'getJSON').and.callFake((cookieName) => testScope.cookieData[cookieName]);
+    spyOn(Cookies, 'get').and.callFake((cookieName) => testScope.cookieData[cookieName]);
     spyOn(Date.prototype, 'toISOString').and.callFake(() => testScope.fakeDatetime);
   });
 
@@ -363,13 +363,13 @@ describe('userPreferencesFactory', () => {
       expect(new Date().toISOString()).toEqual(123);
     });
     it('cookies should be settable and gettable', () => {
-      expect(Cookies.getJSON('abcdef')).toBeUndefined();
+      expect(Cookies.get('abcdef')).toBeUndefined();
       Cookies.set('abcdef', 'this is my cookie value');
-      expect(Cookies.getJSON('abcdef')).toBe('this is my cookie value');
+      expect(Cookies.get('abcdef')).toBe('this is my cookie value');
     });
     it('cookies should be settable with an object', () => {
-      Cookies.set('abcdef', { thisIs: 'AnObject' });
-      expect(Cookies.getJSON('abcdef')).toEqual({ thisIs: 'AnObject' });
+      Cookies.set('abcdef', JSON.stringify({ thisIs: 'AnObject' }));
+      expect(JSON.parse(Cookies.get('abcdef'))).toEqual({ thisIs: 'AnObject' });
     });
   });
 });
--- a/js/test/domain/userPreferencesFactory.spec.ts
+++ b/js/test/domain/userPreferencesFactory.spec.ts
@@ -22,11 +22,11 @@ describe('userPreferencesFactory', () => {
   });
 
   const setConsentCookie = (obj) => {
-    testScope.cookieData.userConsent = { version: '2021.1', ...obj };
+    testScope.cookieData.userConsent = JSON.stringify({ version: '2021.1', ...obj });
   };
 
   const expectTrackingPreferenceToHaveBeenSetWith = (preference) => {
-    expect(Cookies.set).toHaveBeenCalledWith('userConsent', preference, expect.anything());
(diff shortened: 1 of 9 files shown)


js/test/domain/userPreferencesFactory.spec.ts
```diff
@@ -363,13 +363,13 @@ describe('userPreferencesFactory', () => {
...
       Cookies.set('abcdef', 'this is my cookie value');
-      expect(Cookies.getJSON('abcdef')).toBe('this is my cookie value');
+      expect(Cookies.get('abcdef')).toBe('this is my cookie value');
     });
     it('cookies should be settable with an object', () => {
-      Cookies.set('abcdef', { thisIs: 'AnObject' });
-      expect(Cookies.getJSON('abcdef')).toEqual({ thisIs: 'AnObject' });
+      Cookies.set('abcdef', JSON.stringify({ thisIs: 'AnObject' }));
+      expect(JSON.parse(Cookies.get('abcdef'))).toEqual({ thisIs: 'AnObject' });
     });
```
oscarduignan, line 373: not really sure why we have these tests historically

https://github.com/hmrc/tracking-consent-frontend/blame/7f52666b621f37304bc8545e759d29f03f105da9/js/test/domain/userPreferencesFactory.spec.ts#L365-L374

before each test we mock Cookies.set / Cookies.get[JSON], so we are testing our mocks work?

oscarduignan: looking at the date one, that seems to be what we're doing, these are just making sure the fake stuff behaves how we expect, so the JSON.stringify / parse stuff here isn't needed

js/test/domain/userPreferencesFactory.spec.ts
oscarduignan, line 1: We have got the intent right in the PR description, but we don't seem to actually be implementing the JSON.stringify / JSON.parse in the usePreferenceFactory.ts

The indicator is that we haven't had to change the test assertions in this file - because we are stubbing js-cookie

So before we just passed objects to js-cookie - and that's what is tested here

But now we need to be passing JSON.stringify'ed objects

so for example:
```
expect(Cookies.set).toHaveBeenCalledWith('userConsent', {
        version: '2021.1',
        datetimeSet: testScope.fakeDatetime,
        preferences: {
          measurement: true,
          settings: true,
        },
      }, { sameSite: 'strict', expires: 365 });
```
should actually now be
```
expect(Cookies.set).toHaveBeenCalledWith('userConsent', JSON.stringify({
        version: '2021.1',
        datetimeSet: testScope.fakeDatetime,
        preferences: {
          measurement: true,
          settings: true,
        },
      }), { sameSite: 'strict', expires: 365 });
```
if we change the validateCookie from
```
  const validateCookie = (): Cookie | undefined => {
    const cookie = Cookies.getJSON(COOKIE_CONSENT);
    if (cookie === null || cookie === undefined || cookie.preferences === undefined) {
      return undefined;
    }
    const { version } = cookie;
    if (version !== COOKIE_VERSION) {
      return undefined;
    }
    return cookie;
  };
  ```
  to something like
  ```
    const validateCookie = (): Cookie | undefined => {
    try {
      const cookie = Cookies.get(JSON.parse(COOKIE_CONSENT));
      if (cookie === null || cookie === undefined || cookie.preferences === undefined) {
        return undefined;
      }

      const { version } = cookie;
      if (version !== COOKIE_VERSION) {
        return undefined;
      }

      return cookie;
    } catch (e) {
      return undefined;
    }
  };
  ```
  and storePreferences from
  ```
  const storePreferences = (preferences: Preferences) => {
    Cookies.set(COOKIE_CONSENT,
      {
        version: COOKIE_VERSION,
        datetimeSet: new Date().toISOString(),
        preferences,
      }, {
        sameSite: 'strict',
        expires: 365,
      });
  };
  ```
  to
  ```
    const storePreferences = (preferences: Preferences) => {
    Cookies.set(COOKIE_CONSENT,
      JSON.stringify({
        version: COOKIE_VERSION,
        datetimeSet: new Date().toISOString(),
        preferences,
      }), {
        sameSite: 'strict',
        expires: 365,
      });
  };
  ```
  then we should start to see tests in this file fail
  
 if you `sbt clean run` this then open http://localhost:12345/tracking-consent and accept the cookies then the look in the browser dev tools you will see the userConsent is stored as `[object%20Object]` rather than the actual json
 
 I don't know why tests are failing for Tim and on CI - maybe running few more tests somehow? Really odd - because on a clean checkout the tests (test it:test) don't fail for me either - so really confused why it does fail for some, but good that it does

oscarduignan: oh, I think the reproducibility issue might be that we're not running all the tests

https://github.com/hmrc/build-jobs/blob/main/jobs/live/platui.groovy#L425

looks like we have acceptance:test scope that does tests in the browser, bet Tim is running those tests and we aren't

`sbt acceptance:test`

oscarduignan: yup, running that for me:

CleanShot 2024-06-04 at 12 14 33@2x

oscarduignan (approved):

PR #280, opened 2024-06-13
PLATUI-3099 update to Scala 3

TimothyFothergill:
# Purpose of PR
To update to Scala 3, this PR includes:

- Updating the version of scala to `3.3.3`
- Bump:
  - `ui-test-runner` -> `0.31.0`
  - `bootstrap version` -> `9.0.0`
  - `sbt-accessibility-linter` -> `1.0.0`
  - `sbt-auto-build` -> `3.22.0`
  - `scalafmt` -> `3.7.15`
    - add a `runner.dialect` for scala3
  - `play-frontend-hmrc` -> `10.1.0`
- Routes to explicitly use parenthesis
- Updating `implicits` to `given`s and `using`s where appropriate
- Updating wildcard from `._` to `.*`
- `[_]` -> `[?]`
- Exclusion added for /non-existent-page for now as appears to be the same as all other error pages - Tech debt?


accessibility-assessment.json
```diff
@@ -3,6 +3,10 @@
     {
       "path": "/test-only",
       "reason": "Test only endpoints do not need accessibility testing as it is used purely for testing purposes."
+    },
+    {
+      "path": "/non-existent-page",
+      "reason": "This path is used to check for a non-existent page, but many status errors would fail this check as they have the same elements"
     }
   ]
 }
```
oscarduignan, line 12: I don't think we need to do now, but I think test only routes are excluded by default now so might not need that, and then wonder if we could remove the need to have non existent page excluded too if we prefixed the url with test only too

app/uk/gov/hmrc/trackingconsentfrontend/config/ErrorHandler.scala
```diff
@@ -16,21 +16,23 @@
...
 ) extends FrontendErrorHandler {
-  override def standardErrorTemplate(pageTitle: String, heading: String, message: String)(implicit
-    request: Request[_]
-  ): Html =
-    errorTemplate(pageTitle, heading, message)
+
+  override def standardErrorTemplate(pageTitle: String, heading: String, message: String)(using
+    request: RequestHeader
+  ): Future[Html] = Future {
+    errorPage(pageTitle, heading, message)
```
oscarduignan, line 36: I think we should use `Future.successful` here instead because we don't actually need to do the rendering of the error page async

js/test/backstop/public/assets-frontend-3.12.0/javascripts/application.min.js
oscarduignan, line 1: what changed in this file, auto format from editor?

oscarduignan: guess it's just test file :shrug:

project/Webpack.scala
oscarduignan, line 1: TIL https://www.playframework.com/documentation/3.0.x/sbtCookbook#Hooking-into-Plays-dev-mode

oscarduignan (approved):

PR #284, opened 2024-06-19
PLATUI-3053: Moved local help-frontend base URL into local.conf

JoPintoPaul:


app/uk/gov/hmrc/trackingconsentfrontend/config/AppConfig.scala
```diff
@@ -33,8 +33,11 @@ class AppConfig @Inject() (config: Configuration, servicesConfig: ServicesConfig
   private val platformHost: Option[String] =
     config.getOptional[String]("platform.frontend.host")
 
+  private val localHelpFrontendBaseUrl =
+    config.getOptional[String]("help-frontend.base-url-local-testing")
+
   private val helpFrontendHost: String =
-    platformHost.map(_ => "").orElse(config.getOptional[String]("help-frontend.host")).getOrElse("")
+    platformHost.orElse(localHelpFrontendBaseUrl).getOrElse("")
```
oscarduignan, line 40: do you know why we only had a root relative path before rather than using platformHost? why we mapped it to "" - seems odd to do that and I'm wondering if there was some undocumented important reason for it or not

oscarduignan: I wonder if this would impact tracking consent frontend pages viewed on api platform - if after this - api platform pages that linked to pages on api platform, would start linking to www.tax.service.gov.uk urls

JoPintoPaul: Oh yes, good spot. That's probably why! I'll change it.

app/uk/gov/hmrc/trackingconsentfrontend/config/AppConfig.scala
```diff
@@ -33,8 +33,14 @@ class AppConfig @Inject() (config: Configuration, servicesConfig: ServicesConfig
...
-  private val helpFrontendHost: String =
-    platformHost.map(_ => "").orElse(config.getOptional[String]("help-frontend.host")).getOrElse("")
+  private val localHelpFrontendBaseUrl =
+    config.getOptional[String]("help-frontend.base-url-local-testing")
+
+  private val helpFrontendHost: String = platformHost match {
+    // if on a environment with Platform host set, use relative URLs
+    case Some(host) => ""
+    case _          => localHelpFrontendBaseUrl.getOrElse("")
+  }
```
oscarduignan, line 43: and saying it out loud for myself mostly, having the config in local.conf isn't going to impact any things testing with service manager, because no-one should be spinning up help frontend with service manager (did a quick search and can't see anyone doing that)

oscarduignan (approved):

app/uk/gov/hmrc/trackingconsentfrontend/config/AppConfig.scala
oscarduignan: might be worth documenting why we do it - we think - because of api platform

oscarduignan (approved):

PR #296, opened 2024-09-04
PLATUI-3225: Removed deprecated IntegrationTest scope

JoPintoPaul:


it/test/AuditISpec.scala
oscarduignan, line 1: @JoPintoPaul it's logged this as an add rather than move, I think there might still be copies of these in test/it

JoPintoPaul: @oscarduignan Just spotted the same and deleted the Slack message, too speedy! Pushed up correction now.

oscarduignan (approved):

PR #312, opened 2025-03-03
PLATUI-3509 Filter to change cookies back to httpOnly false

PatrykRudnicki:


TimothyFothergill (dismissed):
nice

it/test/CookieConsentFilterISpec.scala
```diff
@@ -0,0 +1,69 @@
...
+
+      val incorrectCookie = Cookie(name = "mdtpurr", value = "cookieVal", httpOnly = true)
+      val expectedCookie = Cookie(name = "mdtpurr", value = "cookieVal", maxAge = Some(31556926), httpOnly = false)
+      val request = FakeRequest(GET, "/tracking-consent/cookie-settings").withCookies(incorrectCookie)
+      val result = route(appWithConfiguration(configuration), request).get
+
+      cookies(result).get("mdtpurr").get shouldBe expectedCookie
+    }
+  }
+}
```
JoPintoPaul, line 68: Should we also have a test that other cookies aren't affected?

test/unit/filters/CookieConsentFilterSpec.scala
```diff
@@ -0,0 +1,85 @@
...
+        val expectedCookie          = Cookie(name = "mdtpurr", value = "cookieVal", maxAge = Some(31556926), httpOnly = false)
+        val request                 = FakeRequest().withCookies(incorrectCookie)
+        val action: EssentialAction = Action(_ => Ok("success"))
+
+        val result = cookieConsentFilter.apply(action)(request).run()
+
+        cookies(result).head mustBe expectedCookie
+      }
+    }
+  }
```
JoPintoPaul, line 84: See comment above - feels like we're missing a test that other cookies are unchanged.

PatrykRudnicki: Added additional tests

TimothyFothergill (dismissed):
lgtm

app/uk/gov/hmrc/trackingconsentfrontend/filters/CookieConsentFilter.scala
```diff
@@ -0,0 +1,35 @@
...
+import scala.concurrent.ExecutionContext
+
+class CookieConsentFilter @Inject() (implicit ec: ExecutionContext) extends EssentialFilter {
+  // This filter is not intended to stay forever, but rather to alleviate an issue with a bug, as per PLATUI-3500
+  override def apply(next: EssentialAction): EssentialAction = EssentialAction { requestHeader =>
+    next(requestHeader).map { result =>
+      val modifiedCookies = requestHeader.cookies.map { c =>
+        if ((c.name == "userConsent" || c.name == "mdtpurr") && c.httpOnly) {
+          c.copy(httpOnly = false, maxAge = c.maxAge.orElse(Some(31556926)))
+        } else c
```
oscarduignan, line 30:
```suggestion
        if ((c.name == "userConsent" || c.name == "mdtpurr")) {
          c.copy(httpOnly = false)
        } else c
```
I don't think we need these bets because those will just be the default of the Cookie case class, browser doesn't send this metadata to the server

PatrykRudnicki: I am not sure about `maxAge`, this came from the spike results, so just followed that, but about `httpOnly` I believe that we only want to fix the broken cookies not the correct ones

app/uk/gov/hmrc/trackingconsentfrontend/filters/CookieConsentFilter.scala
```diff
@@ -0,0 +1,35 @@
...
+      }
+      result.withCookies(modifiedCookies.toSeq: _*)
```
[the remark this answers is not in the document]
oscarduignan: modifiedCookies here will be everything in request.cookies (not just userConsent and mdtpurr)

we only want to send a Set-Cookie header fro userConsent and mdtpurr (no matter what cookies get sent to us)

because otherwise we'll hit the same bug that bas-gateway-frontend had, of overwriting existing cookies with the default settings for the Cookie case class by doing withCookies(request.cookies)

app/uk/gov/hmrc/trackingconsentfrontend/filters/CookieConsentFilter.scala
JoPintoPaul: @PatrykRudnicki @oscarduignan worth breaking out `31556926` into a val with a friendly name? Something like `val secondsInAYear = 31556926`?

it/test/CookieConsentFilterISpec.scala
```diff
@@ -0,0 +1,76 @@
...
+        "help-frontend.base-url-local-testing" -> "http://localhost:9240"
+      )
+    )
+    .build()
+
+  "Given a running instance of tracking consent frontend, calling any endpoint" should {
+    "switch userConsent cookie to httpOnly false if it's true" in {
+      val incorrectCookie = Cookie(name = "userConsent", value = "cookieVal", httpOnly = true)
+      val expectedCookie = Cookie(name = "userConsent", value = "cookieVal", maxAge = Some(31556926), httpOnly = false)
+      val request = FakeRequest(GET, "/tracking-consent/cookie-settings").withCookies(incorrectCookie)
```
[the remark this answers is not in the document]
PatrykRudnicki: Integration tests has been rewritten

app/uk/gov/hmrc/trackingconsentfrontend/filters/CookieConsentFilter.scala
```diff
@@ -0,0 +1,38 @@
...
+class CookieConsentFilter @Inject() (implicit ec: ExecutionContext) extends EssentialFilter {
+  private val yearInSeconds = 31556926
+
+  // This filter is not intended to stay forever, but rather to alleviate an issue with a bug, as per PLATUI-3509
+  override def apply(next: EssentialAction): EssentialAction = EssentialAction { requestHeader =>
+    next(requestHeader).map { result =>
+      val modifiedCookies =
+        requestHeader.cookies.withFilter(c => c.name == "userConsent" || c.name == "mdtpurr").map { c =>
+          if (c.httpOnly) c.copy(httpOnly = false, maxAge = c.maxAge.orElse(Some(yearInSeconds)))
+          else c
```
oscarduignan, line 32: small one

```suggestion
          c.copy(httpOnly = false, maxAge = Some(yearInSeconds))
```

it/test/CookieConsentFilterISpec.scala
```diff
@@ -0,0 +1,67 @@
...
+      cookie must include("Max-Age=31556926")
+    }
+
+    "does not affect other cookies" in {
+      val response = await(wsClient.url(testUrl).addHttpHeaders(("Cookie", "otherCookie=cookieVal")).get())
+      val cookie = response.headerValues("Set-Cookie").filter(_.contains("otherCookie"))
+
+      cookie mustBe empty
+    }
+  }
```
oscarduignan, line 65: I think it's good to keep this in 🙌

test/unit/filters/CookieConsentFilterSpec.scala
oscarduignan, line 1: I don't know if we need these anymore, doesn't seem like it's adding much because it's the same coverage

and could confuse people in future, because I think FakeRequest passes through the cookies set on it directly to the controller action, so it is possible to construct cookies that won't be possible for the action to receive from a real request

PatrykRudnicki: Sure, happy to remove Unit tests for it :)

oscarduignan (approved):
:perfection: 🙌

PR #317, opened 2025-03-26
PLATUI-3564 Migrate VRT tests to playwright

PatrykRudnicki:


.browserslistrc
```diff
@@ -1,3 +0,0 @@
-> 0.5%
-last 2 versions
-not ie <= 9
```
PatrykRudnicki, line 3: Old browser list

package.json
```diff
@@ -76,5 +78,14 @@
     "stylelint-config-standard": "20.0.0",
     "stylelint-scss": "3.16.0",
     "ts-jest": "^29.1.4"
+  },
+  "browserslist": [
+    "last 2 versions",
+    "ie 8",
+    "ie 9",
+    "iOS 9"
+  ],
```
PatrykRudnicki, line 87: I picked up this list from hmrc-frontend. Please let me know if I should keep it to be aligned or the old list should be the one we should be using.

I commented the old one for visibility

oscarduignan: `> 0.5%` would mean include things that ensure stuff will work (no js error) when browser marketshare is higher than 0.5% - but where I'm actually not totally sure where the %s it's referring to come from

a bit related to https://caniuse.com/

I wonder if it's worth actually keeping the old list, and then having a separate ticket to investigate how we're using browserlist stuff (and I guess maybe babel) just in case we end up accidentally stopping tracking consent working for users of older browsers

PatrykRudnicki: of course, will wait for build to prove that it's working and will push commit with older settings 😄  thanks

oscarduignan (approved):

PR #365, opened 2026-04-24
PLATUI-4263 allow and detect use of service navigation

oscarduignan:
* pages served by tracking consent frontend can have service nav enabled
* cookie banner detects if page has service navigation on it (or it's been requested explicitly via data attribute on script tag) and propagates service nav usage onto the links to cookie settings page


app/uk/gov/hmrc/trackingconsentfrontend/views/components/CookieHead.scala.html
```diff
@@ -24,6 +26,7 @@
   id="tracking-consent-script-tag"
   data-gtm-container="a"
   data-language="@messages.lang.code"
+  @if(serviceNavConfig.forceServiceNavigation) {data-use-service-nav}
```
oscarduignan, line 29: I totally missed in my change to previous major version of play-frontend-hmrc that this still was using -nav rather than -navigation

https://github.com/hmrc/play-frontend-hmrc/blob/v12.32.1/play-frontend-hmrc-play-30/src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/helpers/HmrcTrackingConsentSnippet.scala.html#L32

package.json
oscarduignan, line 1: probably loads more dependencies here that could be updated - but just did the one that was blocking me from building locally and will create a tech debt ticket to come back and update node version then update deps with safety of the new min package age security control

app/uk/gov/hmrc/trackingconsentfrontend/views/components/CookieHead.scala.html
oscarduignan: this is needed here because we load the snippet not through play-frontend-hmrc (can't remember the reason for that though)

PatrykRudnicki (dismissed):
LGTM

PatrykRudnicki (approved):

PR #369, opened 2026-06-22
PLATUI-4313 Introducing min-release-age, ignore scripts and npm dependencies updates

PatrykRudnicki:


project/JavaScriptBuild.scala
```diff
@@ -12,12 +12,16 @@ import sbt.*
...
+  val npmInstallPlaywright  = TaskKey[Int]("npm-install-playwright")
+  val npmInstallPuppeteer     = TaskKey[Int]("npm-install-puppeteer")
   val npmTest     = TaskKey[Int]("npm-test")
   val npmBackstop = TaskKey[Int]("npm-backstop")
   val npmBuild    = TaskKey[Int]("npm-build")
 
   val javaScriptSettings: Seq[Setting[?]] = Seq(
     npmInstall := Npm.npmProcess("npm install failed")(baseDirectory.value, "install"),
+    npmInstallPlaywright := Npm.npmProcess("npm install playwright failed")(baseDirectory.value, "install:playwright"),
+    npmInstallPuppeteer := Npm.npmProcess("npm install puppeteer failed")(baseDirectory.value, "install:puppeteer"),
```
oscarduignan: oh for this it's not that we want to install puppeteer and playwright - playwright is already installed via npm Install because it's listed in the deps - and the npmBuild depedency on npmInstall and npmTest dependency on npmBuild

instead what we want to do is trigger download of browsers - because that normally happens (I think?) on npm install - but with lifecycle scripts disabled that won't happen anymore

oscarduignan: oh wait sorry I misread "npm install playwirght failed" as the command 😮‍💨

oscarduignan: did the install:playwright task already get added before?

oscarduignan: https://github.com/hmrc/hmrc-frontend/blob/main/package.json#L24

oscarduignan: I think npmInstallPlaywright will also need to depend on npmInstall because installing the browsers is executing something in an installed dependency-  but we can't put it after npmTest (which is what triggers npmInstall at the moment) because we need browsers for npmTest - but guess don't want npmTest to always run browser install

PatrykRudnicki: That has been sorted. Thank you Oscar for help

package.json
```diff
@@ -77,7 +78,6 @@
   },
   "overrides": {
     "serialize-javascript": "^7.0.3",
-    "playwright": "$playwright-core",
```
oscarduignan, line 80: I wasn't sure if this would stop backstop tests running locally / on jenkins but for me it works in both 🙌  so not sure why we needed the override now to align the direct dep and transitive one

oscarduignan (approved):

